PR #5, opened 2024-07-04
Add compose.yml to run the app locally

wozniakpl:
AB#206960

### DIFF
--- a/compose.yml
+++ b/compose.yml
@@ -1,0 +1,23 @@
+services:
+  backend:
+    build:
+      context: .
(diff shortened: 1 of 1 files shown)


wozniakpl:
@saxix could you please provide instruction on how to use that Dockerfile?
I couldn't set it up using the Makefile present in docker dir and using the image directly as in the compose file provided ends with this:
```
backend-1  | /usr/local/bin/docker-entrypoint.sh: 16: exec: gosu: Exec format error
```

PR #55, opened 2025-03-18
AB#241073: Import from Kobo: filter project

sergey-misuk-valor:
# Legal Boilerplate

Look, I get it.
under HOPE's choice of terms.


ntrncic (approved):

src/country_workspace/contrib/kobo/forms.py
```diff
@@ -3,12 +3,6 @@
 
 class ImportKoboForm(forms.Form):
     batch_name = forms.CharField(required=False, help_text="Label for this batch")
-    country_code = forms.CharField(
-        required=False,
```
saxix, line 7: Have you checked how it is managed in HOPE? I'm not sure we are relying on country code right now.

src/country_workspace/config/fragments/app.py
```diff
@@ -5,6 +5,7 @@
 
 TENANT_TENANT_MODEL = "country_workspace.Office"
 TENANT_HQ = "= HQ ="
+TENANT_HQ_COUNTRY = "AFG"
```
domdinicola, line 8: what's this settings for? not sure we need to hardcode AFG anywhere

src/country_workspace/models/office.py
```diff
@@ -12,6 +12,7 @@ class Office(BaseModel):
     code = models.CharField(max_length=100, blank=True, null=True, db_index=True, unique=True)
     slug = models.SlugField(max_length=100, blank=True, null=True, db_index=True, unique=True)
     active = models.BooleanField(default=False)
+    country_iso_code = models.CharField(max_length=3, blank=False, null=False)
```
domdinicola, line 15: I would just do the field not mandatory

src/country_workspace/config/fragments/app.py
sergey-misuk-valor: For mandatory country_iso_code in HQ office

src/country_workspace/models/office.py
sergey-misuk-valor: What should client do when it's empty? Fetch all projects?

src/country_workspace/contrib/kobo/forms.py
saxix: not sure about this. Please add a note here about how this will work with Kobo, how we will manage existing forms and what changes are needed in the various systems. HOPE BusinessArea does not represent a country, and it does not have any country attribute. (cc @ntrncic, @domdinicola).

src/country_workspace/config/fragments/app.py
saxix: this looks like a bad side effect of a wrong decision to use country code in BA

domdinicola (approved):

PR #145, opened 2025-06-19
Bugfix/sentry id not generating

patsatsia:
When an exception is captured by the logger's Sentry integration, subsequent sentry_sdk.capture_exception() calls on the same exception object don't generate new IDs, leaving sentry_id empty


domdinicola:
rule: if you raise log error if not log exception

domdinicola (approved):

PR #146, opened 2025-06-20
Fix issue of failing 'validate_program' task function by adding state…

patsatsia:


saxix (changes requested):
Please add to the description the error we are trying to fix and why this patch should fix it. 
I suppose Program should be set in the task

patsatsia:
> Please add to the description the error we are trying to fix and why this patch should fix it. I suppose Program should be set in the task

it is trying to resolve this [issue](https://monitoring.hope.unicef.org/organizations/hope/issues/13693/?query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=12)

It is coming from AdminLevelChoice initialization (src/country_workspace/contrib/hope/geo.py)
Tenant is not set to the state within the celery task. I have tried to run synchronous function and it worked just fine 
image

PR #160, opened 2025-07-04
Feature: Program Serializer

patsatsia:


domdinicola:
@saxix dukpy what do you think about this?

tests/models/test_m_serializer.py
```diff
@@ -0,0 +1,31 @@
...
+    assert str(data_serializer) == f"DataSerializer: {data_serializer.name}"
+
+
+def test_clean_valid(data_serializer: DataSerializer):
+    data_serializer.clean()
+
+
+@pytest.mark.parametrize("code", ["", "function {"])
+def test_clean_invalid_js(data_serializer: DataSerializer, code: str):
+    data_serializer.code = code
```
domdinicola, line 29: let's add a tests with "documents" and "accounts"

src/country_workspace/contrib/hope/geo.py
```diff
@@ -71,6 +71,9 @@ def validate_with_parent(self, parent_value: Any, value: Any) -> None:
         if parent_value and self.prepare_value(value) not in choices:
             raise ValidationError("Not valid child for selected parent")
 
+    def validate(self, value: Any) -> None:
```
domdinicola, line 74: why this is needed?

src/country_workspace/contrib/hope/constants.py
```diff
@@ -13,3 +13,13 @@
 ADMINAREAS_FIELDSET_NAME: Final[str] = "HOPE Admin Areas"
 ACCOUNT_FIELDSET_NAME: Final[str] = "HOPE Account"
 DOCUMENT_FIELDSET_NAME: Final[str] = "HOPE Document"
+
+DOCUMENT_PREFIXES_TO_TYPE_MAPPING = {
+    "national_id__": "national_id",
+    "national_passport__": "national_passport",
+}
```
domdinicola, line 20: maybe we should have a dynamic functionality to remove trailing chars like spaces, _, - :

src/country_workspace/contrib/hope/push.py
```diff
@@ -188,18 +191,39 @@ def safe_post(self, path: str, data: Any, error_msg: str) -> dict[str, Any] | No
             self.total["errors"].append(f"{error_msg}: {e}")
             return None
 
+    @cached_property
+    def program(self) -> Program:
+        return Program.objects.get(hope_id=self.program_hope_id)
+
+    @staticmethod
+    def _set_types(item: Validable) -> None:
+        for prefix in DOCUMENT_PREFIXES_TO_TYPE_MAPPING:
```
domdinicola, line 200: would be good to find a way to not iterate on prefixes..

src/country_workspace/migrations/0021_merge_20250710_1203.py
```diff
@@ -0,0 +1,12 @@
+# Generated by Django 5.2.3 on 2025-07-10 12:03
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("country_workspace", "0019_mappingimporter"),
```
domdinicola, line 8: let's avoid merge migrations, delete and regenerate

src/country_workspace/models/mixins.py
```diff
@@ -0,0 +1,37 @@
+from collections import defaultdict
+
+
+class FlexFieldGroupingMixin:
+    def get_grouping_info(self) -> dict[str, list[str]]:
+        grouping_info = defaultdict(list)
+        for member in self.checker.members.select_related("fieldset").all():
+            if not member.prefix:
+                continue
+            if member.group == "":
```
domdinicola, line 10: this can be done with an or

src/country_workspace/contrib/hope/constants.py
```diff
@@ -13,3 +13,13 @@
 ADMINAREAS_FIELDSET_NAME: Final[str] = "HOPE Admin Areas"
 ACCOUNT_FIELDSET_NAME: Final[str] = "HOPE Account"
 DOCUMENT_FIELDSET_NAME: Final[str] = "HOPE Document"
+
+DOCUMENT_PREFIXES_TO_TYPE_MAPPING = {
+    "national_id__": "national_id",
+    "national_passport__": "national_passport",
```
domdinicola, line 19: let's change the double __ with a single _. Since the XLSX template is using just one.

src/country_workspace/contrib/hope/constants.py
datamik: The prefix and the type are not always the same, for example we have prefix **phone__** and the closest type i have found in the account_types table was **mobile**. 
If i picked the wrong one and they will be always the same, of course we can make it dynamic.

src/country_workspace/contrib/hope/geo.py
patsatsia: within the latest hope-flex-fields version, there is a change that requires validate method to be declared explicitly 
https://github.com/unicef/hope-flex-fields/blob/develop/src/hope_flex_fields/mixin.py

PR #167, opened 2025-07-10
change sort in admin

vitali-yanushchyk-valor:
AB#261279


domdinicola:
@vitali-yanushchyk-valor not just in the admin, also in the app!

PR #169, opened 2025-07-14
fix ! format values for XLSX export: lists/tuples as comma-separated

vitali-yanushchyk-valor:
[AB#263918](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/263918)

### DIFF
--- a/src/country_workspace/workspaces/admin/cleaners/bulk_update.py
+++ b/src/country_workspace/workspaces/admin/cleaners/bulk_update.py
@@ -161,9 +161,10 @@ def create_xls_importer(
             worksheet.write(0, i, fld_name, header_format)
     worksheet.freeze_panes(1, 0)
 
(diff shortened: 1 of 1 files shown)


domdinicola (approved):

src/country_workspace/workspaces/admin/cleaners/bulk_update.py
```diff
@@ -161,9 +161,10 @@ def create_xls_importer(
             worksheet.write(0, i, fld_name, header_format)
     worksheet.freeze_panes(1, 0)
 
+    fmt = lambda v: ", ".join(map(str, v)) if isinstance(v, list | tuple) else str(v if v is not None else "")
```
sergey-misuk-valor, line 164: It's better to have a regular named function here. The only difference with a regular named function here is no typing and less readable code

PR #176, opened 2025-07-22
Simple solution for import mode specific errors messages

datamik:
[#265156](https://dev.azure.com/unicef/ICTD-HCT-MIS/_workitems/edit/265156)


src/country_workspace/workspaces/exceptions.py
```diff
@@ -1,8 +1,8 @@
 class BeneficiaryValidationError(Exception):
-    def __init__(self, beneficiary: str, key: int) -> None:
-        super().__init__(beneficiary, key)
-        self.beneficiary = beneficiary
```
domdinicola, line 4: this are good to remove?

PR #185, opened 2025-07-25
use latest version hope-flex-fields library (0.7.2)

vitali-yanushchyk-valor:
AB#265535


saxix (changes requested):
why pyproject.toml is not in the PR ?

vitali-yanushchyk-valor:
> why pyproject.toml is not in the PR ?

because pyproject.toml has https://github.com/unicef/hope-country-workspace/blob/266e4349ad96e32f28677f3fb2fa351358bbecd7/pyproject.toml#L52
so for hope-flex-fields version 0.7.2 upgrade uv.lock was enough

saxix (changes requested):
Iif we upgraded hope-flex-field to 0.7.2, because wee need the fixes/updates from 0.7.2, pyproject.toml must contains ">=0.7.2"

PR #188, opened 2025-07-30
Store cleaned data and raw data when the excell file is imported.

datamik:
Update flex_fields with cleaned_data on every validation
https://dev.azure.com/unicef/ICTD-HCT-MIS/_workitems/edit/265154

# Legal Boilerplate

Look, I get it.
under HOPE's choice of terms.


domdinicola (approved):

domdinicola (approved):

src/country_workspace/datasources/rdi.py
```diff
@@ -106,10 +106,11 @@ def process_households(sheet: Sheet, job: AsyncJob, batch: Batch, config: Config
         try:
             mapping[household_key] = cast(
                 "Household",
-                job.program.households.create(
-                    batch=batch,
-                    name=label,
+                Household.objects.create(
```
vitali-yanushchyk-valor, line 109: Maybe it will looks better?
```                
batch.household_set.create(
      name=str(label),
      flex_fields=flex_fields,
      raw_data=flex_fields,
),
```

vitali-yanushchyk-valor (approved):

src/country_workspace/datasources/rdi.py
datamik: The issue with the previous approach was that all Household instances shared the same checker instance, which made the design more error-prone. I checked, and the same applies to the current solution you proposed. Given that, I think it’s safer to stick with the existing implementation for now.

PR #202, opened 2025-08-29
Add id fields to eliminate alien fields issue when importing individuals and people

datamik:
https://dev.azure.com/unicef/ICTD-HCT-MIS/_workitems/edit/273244


datamik:
> @datamik pls check tests

Test seems to be OK

PR #339, opened 2026-02-24
298132: Add collision detection for identity fields in import and val…

arsen-vs:
…idation processes

- Implemented `detect_and_mark_collisions_for_batch` to mark incoming records with identity collisions during batch import.
- Added `check_identity_collision` to validate identity fields in `Household` and `Individual` models, updating errors if collisions are found.
- Updated `import_from_rdi` and `validate_with_checker` methods to incorporate collision checks, ensuring data integrity during import and validation.

[AB#298132](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/298132)


datamik (changes requested):
Do we need to check for collisions on every validation, or only during import? Should the identity_field be editable?

Should we check for collisions within the batch being imported? If I understand correctly, we only check the batch against existing data, not within the batch itself.

domdinicola (approved):

tests/models/test_m_collision.py
```diff
@@ -0,0 +1,354 @@
...
+    hh.save(update_fields=["errors"])
+
+    detect_and_mark_collisions_for_batch(batch)
+
+    hh.refresh_from_db()
+    assert "identity" not in hh.errors
+
+
+def test_detect_and_mark_skips_empty_identity_values():
+    """Records with blank identity values are ignored."""
```
saxix: Usage of factories inside tests should be limited to extraordinary circumstances. Use fixtures to wrap Factorie

saxix (changes requested):

datamik (approved):

vitali-yanushchyk-valor (approved):

saxix (approved):

domdinicola:
@datamik pls fix conflicts

PR #374, opened 2026-04-30
Improve mass update action performance

datamik:
Avoid loading all date at once;
Defer raw_data field;
Avoid not necessary queries and cache version incr

### DIFF
--- a/src/country_workspace/workspaces/admin/cleaners/mass_update.py
+++ b/src/country_workspace/workspaces/admin/cleaners/mass_update.py
@@ -108,11 +108,12 @@ def mass_update_impl(
     config: "FormOperations",
     create_missing_fields: bool = False,
 ) -> None:
-    program = None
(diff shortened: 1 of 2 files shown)


vitali-yanushchyk-valor (approved):

src/country_workspace/workspaces/admin/cleaners/mass_update.py
```diff
@@ -118,4 +122,6 @@ def mass_update_impl(
                 elif create_missing_fields:
                     func = operations.get_function_by_id(op)
                     record.flex_fields[field_name] = func("", new_value)
-            record.save()
+            record.save(update_fields=["flex_fields"])
+    if program:
```
domdinicola: we assume it's always the same program, why we don't get once and for all, instead of doing an if every cycle?

roma-valor (approved):

PR #375, opened 2026-04-30
Improve cleanup_merged_rdp_data performance

datamik:
Clean up heavy fields before deletion to prevent them from being loaded into memory when the post_delete signal triggers.

### DIFF
--- a/src/country_workspace/tasks.py
+++ b/src/country_workspace/tasks.py
@@ -13,7 +13,7 @@ from redis_lock import Lock
 from country_workspace.cache.handlers import suppress_cache_updates
 from country_workspace.cache.manager import cache_manager
 from country_workspace.config.celery import app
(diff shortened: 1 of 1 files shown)


vitali-yanushchyk-valor:
What about add `cache_manager.incr_cache_version()` to the `cleanup_merged_rdp_data()`?

src/country_workspace/tasks.py
```diff
@@ -52,6 +52,8 @@ def cleanup_merged_rdp_data() -> None:
             batch_h_pks = list(households_to_delete.values_list("pk", flat=True)[:batch_size])
```
vitali-yanushchyk-valor: What about add ordering like
`batch_h_pks = list(households_to_delete.order_by("pk").values_list("pk", flat=True)[:batch_size])`

vitali-yanushchyk-valor (approved):

domdinicola (approved):

PR #384, opened 2026-05-07
Fix Invalidation issue when datachecker is changed

datamik:
[AB#314710](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/314710)

**Summary**
Editing DataChecker-related models in Django admin triggered post_save signals that bulk-invalidated every validated household and individual in every program using that checker. A single form save could fire many signals (parent + inlines), so the same programs were invalidated repeatedly.

**Problem**

- Saving DataChecker itself (e.g. name/description) ran full invalidation even though those fields do not affect validation rules.
- Saving Fieldset, FlexField, or DataCheckerFieldset via admin (often several .save() calls per submit) re-ran invalidation for the same programs on each signal.

**Fix**
- Signals — Stop listening to DataChecker post_save. Composition changes are still handled via DataCheckerFieldset (and related models).
- Admin — Override hope_flex_fields admins (DataChecker, Fieldset, FlexField) to wrap POST saves in collect_invalidations(), so all signals in one form submit defer invalidation and each program is processed once at the end. FieldsetAdmin.all_fields is wrapped the same way.

### DIFF
--- a/src/country_workspace/admin/__init__.py
+++ b/src/country_workspace/admin/__init__.py
@@ -6,6 +6,7 @@ from smart_admin.smart_auth.admin import ContentTypeAdmin, PermissionAdmin
 
 from ..cache.smart_panel import panel_cache
 from .batch import BatchAdmin
+from .flex_fields import CWDataCheckerAdmin, CWFieldsetAdmin, CWFlexFieldAdmin
 from .beneficiary_group import BeneficiaryGroupAdmin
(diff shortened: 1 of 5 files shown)


domdinicola:
@datamik resume

vitali-yanushchyk-valor (approved):

roma-valor (approved):

PR #394, opened 2026-05-26
Feature/bitcaster integration

roma-valor:
# Legal Boilerplate
[AB#315502](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/315502)

Adds Bitcaster notification support. When data is imported, validated, or pushed to HOPE, the app sends an async event to Bitcaster via a Celery task.

- New country_workspace.notifications app

- Includes BitcasterClient, Django signals, and signal handlers that enqueue Celery notification tasks.
Integrates with Bitcaster using the bitcaster-sdk.
Uses strict push event contracts:
rdi_push_completed with payload: program_id, pushed_count
rdp_push_status_changed with payload: program_id, rdp_id, status
Bitcaster configuration

- Adds BITCASTER_* settings wired through SmartEnv and settings.py.
Adds BITCASTER_ENABLED feature flag; when disabled, notification tasks log and skip sending.
Signal emission

- data_imported_signal is emitted after successful batch imports from Aurora, Kobo, and RDI sources.
HOPE push orchestration emits:
rdi_push_completed for completed push volume,
rdp_push_status_changed for RDP status transitions (success/failure).
Chunked validation completion tracking

- create_validation_jobs assigns shared validation_run_id and validation_total_chunks to chunk jobs to emit one final validation completion event after all chunks finish.

- Adds/updates tests for notification handlers, Bitcaster client/task behavior, feature-flag behavior, and HOPE push signal payloads/event names.


domdinicola:
@roma-valor pls fix conflicts

src/country_workspace/config/__init__.py
```diff
@@ -197,6 +197,11 @@ class Group(Enum):
     "IGNORE_UNSUPPORTED_FEATURES": (bool, False, True),
     "REQUESTS_TIMEOUT": (int, 30, 30),
     "DEBUG_API_REQUESTS": (bool, False, False),
+    "BITCASTER_API_URL": (str, "", "http://localhost:8002", False, "Bitcaster API URL"),
+    "BITCASTER_API_KEY": (str, "", "", False, "Bitcaster API key"),
+    "BITCASTER_ORGANIZATION_SLUG": (str, "", "organization", False, "Bitcaster organization slug"),
```
saxix: BITCASTER_ORGANIZATION_SLUG = "unicef"
BITCASTER_PROJECT_SLUG = "hope"

saxix (changes requested):

src/country_workspace/config/__init__.py
roma-valor: Fixed

saxix (commented):
- check bitcaster_sdk, it has a client and a retry mechanism, maybe we save some code
- `rdi_pushed_signal` seems too "flexible", this flexibility can increase the filtering cost on BC side, (es. target and rdp_id seem optionals) maybe 2 signals?

ntrncic:
@roma-valor description of PR please.

roma-valor:
@saxix  
- Replaced custom client with `bitcaster-sdk`. 
- Removed `target` param from `rdi_pushed_signal`.

roma-valor:
@saxix resolved your suggestions, can you please take a look?

src/country_workspace/notifications/tasks.py
```diff
@@ -0,0 +1,33 @@
...
+    backend = get_notification_backend()
+    if not getattr(backend, "is_configured", False):
+        logger.warning("Skipping Bitcaster task: client not configured.")
+        return
+
+    try:
+        success = send_notification_event(event_name, payload)
+        if not success:
+            logger.warning("Bitcaster client returned false for event '%s'", event_name)
+    except NotifyError as exc:  # pragma: no cover
```
saxix: do not see send_notification_event raising NotifyError

src/country_workspace/notifications/bitcaster_client.py
```diff
@@ -0,0 +1,51 @@
...
+from typing import Any
+from urllib.parse import urlparse
+
+from bitcaster_sdk.client import Client as SDKClient
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+class BitcasterClient:
```
saxix: check bitcaster-sdk for bitcaster client

saxix (changes requested):

src/country_workspace/notifications/bitcaster_client.py
roma-valor: We already use bitcaster SDKClient, it's just a custom extension for us

src/country_workspace/notifications/tasks.py
roma-valor: Missed to raise it in client.
Resolved

src/country_workspace/notifications/bitcaster_client.py
saxix: I see, let change the name to `BitcasterManager` OR make this client inherit from the original, this will reduce ambiguity

saxix (dismissed):

domdinicola:
@roma-valor pls resolve conflicts

src/country_workspace/workspaces/admin/cleaners/validate.py
```diff
@@ -1,19 +1,75 @@
...
+    validation_run_id: str,
+    total_chunks: int,
+    program_id: int,
+    context: str,
+    valid: int,
+    invalid: int,
+    sender: type[Model],
+) -> None:
+    cache_key = f"validation-run:{validation_run_id}"
+    progress: dict[str, int | str] = cache.get(cache_key, {})
```
vitali-yanushchyk-valor: This cache read-modify-write sequence is not atomic. Validation chunks are queued as independent jobs and may complete concurrently, so multiple workers can read the same progress and overwrite each other’s increments. In that case, `completed_chunks` may never reach `total_chunks`, and the `validation_completed` event may not be emitted.

src/country_workspace/contrib/kobo/sync.py
```diff
@@ -468,11 +469,20 @@ def import_data(job: AsyncJob) -> ImportResult:
...
         job.ensure_not_cancelled(refresh=True)
         # Mark batch complete in a dedicated transaction.
         with transaction.atomic():
             Batch.objects.select_for_update().filter(pk=batch.pk).update(status=Batch.BatchStatus.COMPLETE)
+
+        data_imported_signal.send(
+            sender=Batch,
+            program_id=batch.program_id,
+            batch_id=batch.id,
+            record_count=household_counter + individual_counter,
```
vitali-yanushchyk-valor: Could it be that these counters only include records imported by the current job? For example, if a timeboxed Kobo import processes 1,000 records in the first job and 100 in the final job, would the event report `record_count=100` instead of `1,100`? Should this use the completed batch totals or persist cumulative counts across continuation jobs?

src/country_workspace/notifications/tasks.py
```diff
@@ -0,0 +1,30 @@
...
+    backend = get_notification_backend()
+    if not getattr(backend, "is_configured", False):
+        logger.warning("Skipping Bitcaster task: client not configured.")
+        return
+
+    try:
+        success = send_notification_event(event_name, payload)
+        if not success:
+            logger.warning("Bitcaster client returned false for event '%s'", event_name)
+    except NotifyError as exc:
```
vitali-yanushchyk-valor, line 29: What delivery guarantees are expected for these events? Is it intentional that `NotifyError` is only logged, the task is not retried, and the event may be lost permanently?

src/country_workspace/workspaces/admin/cleaners/validate.py
```diff
@@ -71,16 +150,35 @@ def _validate_and_count(objs: Iterable[Model]) -> tuple[int, int]:
     return valid, invalid
 
 
-def create_validation_jobs(description: str, owner: str, program: Program, queryset: QuerySet) -> AsyncJob:
+def create_validation_jobs(
+    description: str, owner: str, program: Program, queryset: QuerySet, *, context: str = "total"
+) -> AsyncJob | None:
```
vitali-yanushchyk-valor: Could you clarify what `context` represents here? `"rdi"` is also the name of the XLS import source, but it is currently passed to `create_validation_jobs()` for RDI/XLS imports, Kobo imports, and batch reprocessing. `"total"` is used for Validate Entire Programme, while Aurora import validation also falls back to `"total"` by default. The current meaning of this field is therefore unclear.

roma-valor: changed the behavior here, now it's more clear

src/country_workspace/notifications/tasks.py
roma-valor: bitcaster sdk has it's own retry mechanism, so we don't need to add another here.
We log because we don't need to break the flow because notification was not sent.

src/country_workspace/contrib/kobo/sync.py
roma-valor: resolved

src/country_workspace/workspaces/admin/cleaners/validate.py
roma-valor: resolved

src/country_workspace/notifications/signals.py
```diff
@@ -0,0 +1,18 @@
...
+
+# Triggered when validation processing is finished.
+# Expected kwargs: program_id (int), validation_scope (str - "program" or "batch"), results (dict)
+validation_completed_signal = django.dispatch.Signal()
+
+# Triggered when an RDI push cycle successfully completes.
+# Expected kwargs: program_id (int), pushed_count (int)
+rdi_push_completed_signal = django.dispatch.Signal()
+
+# Triggered when an RDP record status changes.
```
vitali-yanushchyk-valor: Could you clarify whether this signal is intended for every RDP status change or only for push results? The comment suggests the former, but currently the signal is emitted only for `SUCCESS` and `FAILURE`; creating an RDP with `PENDING` and later changing it to `CANCELLED` are not reported.

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -268,6 +278,24 @@ def push_existing_rdp_core(job: AsyncJob) -> dict[str, Any]:
             deduplication_set_id=deduplication_set_id,
             processor=hope_processor,
         )
+        pushed_count = (
+            hope_processor.total.get("households", 0)
+            + hope_processor.total.get("individuals", 0)
+            + hope_processor.total.get("people", 0)
+        )
+
+        rdi_push_completed_signal.send(
```
vitali-yanushchyk-valor: With the newly added `rdi_already_merged` path, the record-push steps are skipped, but the outer flow still emits both success events, and `pushed_count` will be `0`. Should this path emit only `rdp_push_status_changed`, since it reconciles the local RDP with an already merged RDI rather than completing a new push?

For example:

```python
if not hope_processor.rdi_already_merged:
    rdi_push_completed_signal.send(...)
```

src/country_workspace/notifications/signals.py
roma-valor: Changed the comment

src/country_workspace/contrib/hope/push/orchestration.py
roma-valor: resolved

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -249,6 +253,12 @@ def push_existing_rdp_core(job: AsyncJob) -> dict[str, Any]:
                         hope_rdi_id=hope_processor.hope_rdi_id or "N/A",
                         is_push_locked=False,
                     )
+                    rdp_push_status_changed_signal.send(
```
vitali-yanushchyk-valor: Could we send this signal only after the transaction commits? Its receiver queues a Celery task, so a broker failure here could roll back the newly saved `FAILURE` status and surface the broker exception instead of the original `HopePushError`. 
Perhaps the signal dispatch could be registered with `transaction.on_commit(..., robust=True)`.

roma-valor: fixed

vitali-yanushchyk-valor (approved):

arsen-vs (approved):

PR #395, opened 2026-05-28
Add functionality to run formulas on existing records

arsen-vs:
- Introduced a new form, `RunTransformerForm`, to allow users to select a batch and specify the record type for formula application.
- Implemented a new admin action, `run_on_existing_records`, to handle the execution of formulas directly in the Country Workspace.
- Created a corresponding HTML template for the form, enhancing the user interface for running formulas without rule commits.
- Added permission checks to ensure users can only run formulas on batches they are authorized to process.

[AB#289969](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/289969)

### DIFF
--- a/src/country_workspace/workspaces/admin/batch/reprocessing.py
+++ b/src/country_workspace/workspaces/admin/batch/reprocessing.py
@@ -16,7 +16,8 @@ from country_workspace.contrib.kobo.sync import (
     build_household_processor as build_kobo_household_processor,
     build_individual_processor as build_kobo_individual_processor,
 )
-from country_workspace.models import AsyncJob, Batch, Household, Individual, MappingImporter, Program
+from country_workspace.models import AsyncJob, Batch, Household, Individual, MappingImporter, Program, Transformer
+from country_workspace.utils.collector_linkage import sync_collector_links
 from country_workspace.utils.fields import to_reference_key
 from country_workspace.utils.import_flow import run_batch_postprocessing, build_import_processor
 from country_workspace.workspaces.admin.cleaners.validate import create_validation_jobs
@@ -53,6 +54,26 @@ def _preserve_flex_fields(
     ), preserved
 
 
+def _get_batch_from_job(job: AsyncJob) -> tuple[int, Batch]:
+    batch_id = job.config.get("batch_id")
+    if not batch_id:
+        raise ValueError("batch_id is required in job config")
+
+    batch = Batch.objects.select_related("program", "country_office").filter(pk=batch_id).first()
+    if not batch:
+        logger.error("Batch %s not found", batch_id)
+        raise Batch.DoesNotExist(f"Batch {batch_id} not found")
+    return batch_id, batch
+
+
+def _sync_household_refs(batch: Batch) -> None:
+    match batch.source:
+        case Batch.BatchSource.KOBO:
+            _sync_kobo_household_refs(batch)
+        case Batch.BatchSource.RDI:
+            _sync_rdi_household_refs(batch)
+
+
 def _apply_import_processor(
     record: Household | Individual,
     processor: Callable[[Any], dict[str, Any]],
@@ -71,6 +92,17 @@ def _apply_import_processor(
     record.last_checked = None
     record.errors = {}
     record.save(update_fields=["flex_fields", "last_checked", "errors"])
+    return True
+
+
+def _apply_transformer_only(record: Household | Individual, transformer: Transformer) -> bool:
+    data = record.flex_fields if isinstance(record.flex_fields, dict) else {}
+    transformed = transformer.apply(data.copy())
+    if transformed == record.flex_fields:
+        return False
+
+    record.flex_fields = transformed
+    record.save(update_fields=["flex_fields"])
     return True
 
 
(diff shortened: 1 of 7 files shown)


src/country_workspace/workspaces/admin/transformer.py
```diff
@@ -63,6 +119,72 @@ def delete_queryset(self, request: HttpRequest, queryset: QuerySet[CountryTransf
         super().delete_queryset(request, queryset)
         self._invalidate_transformer_cache()
 
+    @button(
+        label="Run Formula on Existing Records",
+        change_form=True,
+        html_attrs={"title": "Run this formula in Country Workspace without rule commits"},
```
vitali-yanushchyk-valor, line 143: What about add decorator-level `permission` for button.

vitali-yanushchyk-valor:
`ApplyTo` controls which records receive the transformer. However, this action runs the full `reprocess_batch` workflow, so both households and individuals are still rebuilt from `raw_data` and have their validation state reset.
Could we use `apply_batch_transformers()` here instead, so that only the selected records are transformed?

src/country_workspace/workspaces/admin/transformer.py
arsen-vs: Solved

roma-valor (approved):

src/country_workspace/workspaces/admin/batch/reprocessing.py
```diff
@@ -378,3 +386,38 @@ def reprocess_batch(job: AsyncJob) -> dict[str, Any]:
...
+    )
+    individual_transformer_id, _ = _resolve_config_object(
+        batch.country_office.transformers.all(),
+        job.config.get("individual_transformer_id"),
+        "Individual transformer",
+    )
+    if not household_transformer_id and not individual_transformer_id:
+        raise ValueError("At least one transformer id is required in job config")
+
+    postprocessing_result = run_batch_postprocessing(
```
vitali-yanushchyk-valor: Do we intentionally run the full postprocessing flow here, including household reference sync and collector links?

If this action is supposed to only apply the selected formula, could we call the transformer utility (`utils.import_flow.transformations.apply_batch_transformers`) directly instead of `run_batch_postprocessing()`?

src/country_workspace/workspaces/admin/transformer.py
```diff
@@ -1,15 +1,71 @@
...
+        queryset=Batch.objects.none(),
+        label=_("Batch"),
+        help_text=_("Select an existing batch to update records before pushing to HOPE."),
+    )
+    apply_to = forms.ChoiceField(
+        label=_("Apply formula to"),
+        choices=ApplyToOptions.choices,
+        help_text=_("Choose which record type should be updated by this formula."),
+    )
+
```
srugano, line 37: If the form is ever initialized without a specific program context ( program=None ), the target options will include  Households .  If a user selects a batch belonging to a non-master-detail program and chooses "Households only", both transformer IDs in the queued job configuration will end up being  None , causing the background task to fail with a  ValueError . @arsen-vs 

  We can prevent this by adding a validation in the form's  clean  method:

```python
    def clean(self) -> dict[str, Any]:                                                                                     
        cleaned_data = super().clean()                                                                                     
        batch = cleaned_data.get("batch")                                                                                  
        apply_to = cleaned_data.get("apply_to")                                                                            
                                                                                                                           
        if batch and apply_to:                                                                                             
            if not batch.program.is_master_detail and apply_to == self.ApplyToOptions.HOUSEHOLDS:                          
                self.add_error(                                                                                            
                    "apply_to",                                                                                            
                    _("Cannot apply formula to households on a program that is not master-detail.")                        
                )                                                                                                          
        return cleaned_data 
```
what do you think? no need to start a failing job

src/country_workspace/workspaces/admin/batch/reprocessing.py
```diff
@@ -378,3 +389,37 @@ def reprocess_batch(job: AsyncJob) -> dict[str, Any]:
...
+    )
+    if not household_transformer_id and not individual_transformer_id:
+        raise ValueError("At least one transformer id is required in job config")
+
+    transformer_result = apply_transformers_to_batch(
+        batch,
+        household_transformer_id=household_transformer_id,
+        individual_transformer_id=individual_transformer_id,
+    )
+
```
vitali-yanushchyk-valor, line 415: Should we trigger validation jobs here after applying the transformer or not?

src/country_workspace/workspaces/admin/transformer.py
```diff
@@ -1,15 +1,84 @@
...
+            self.fields["apply_to"].choices = [
+                (self.ApplyToOptions.INDIVIDUALS, self.ApplyToOptions.INDIVIDUALS.label),
+            ]
+
+    def clean(self) -> dict[str, object]:
+        cleaned_data = super().clean()
+        batch = cleaned_data.get("batch")
+        apply_to = cleaned_data.get("apply_to")
+
+        if batch and apply_to == self.ApplyToOptions.HOUSEHOLDS and not batch.program.is_master_detail:
```
vitali-yanushchyk-valor: Should this also reject BOTH for non-master-detail batches? When no program is selected, the form allows BOTH, but for a non-master-detail batch this later creates only individual_transformer_id, so the selected "Households and Individuals" option is silently reduced to individuals only.

srugano (approved):

vitali-yanushchyk-valor (approved):

PR #396, opened 2026-05-28
Add Steficon formula to Country Workspace

arsen-vs:
289969: Add engine field to Transformer model and update related functionality

- Introduced a new `engine` field in the `Transformer` model to specify the formula engine (JavaScript or Steficon).
- Updated the `TransformerAdmin` and `CountryTransformerAdmin` to include the `engine` field in the admin interface.
- Implemented validation for Steficon formulas in the `clean` method of the `Transformer` model.
- Created a new `SteficonExecutor` class to handle the execution of Steficon-style formulas.
- Added unit tests for the new functionality, ensuring correct behavior for both JavaScript and Steficon engines.

### DIFF
--- a/src/country_workspace/models/transformer.py
+++ b/src/country_workspace/models/transformer.py
@@ -1,17 +1,23 @@
 import logging
 from typing import Any
 from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
 from country_workspace.models.base import BaseModel
 from country_workspace.utils.js_executor import JavaScriptExecutor
-from country_workspace.validators.mapping import ValueTransformationRulesValidator
+from country_workspace.utils.steficon_executor import SteficonExecutor
+from country_workspace.validators.mapping import SteficonTransformationRulesValidator, ValueTransformationRulesValidator
 
 logger = logging.getLogger(__name__)
 
 
 class Transformer(BaseModel):
+    class Engine(models.TextChoices):
+        JAVASCRIPT = "JAVASCRIPT", _("JavaScript")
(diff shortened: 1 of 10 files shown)


src/country_workspace/utils/steficon_executor.py
```diff
@@ -0,0 +1,93 @@
...
+
+class SteficonExecutor:
+    """Execute Steficon-style Python formula code.
+
+    Formula receives:
+      - context: dict with "record"
+      - result: object exposing "value" and "extra"
+    """
+
+    FORBIDDEN_TOKENS = (
```
saxix, line 29: `FORBIDDEN_PATTERNS` seems more appropriate

src/country_workspace/models/transformer.py
```diff
@@ -42,13 +54,34 @@ class Meta:
...
+            raise ValidationError({"value_transformations": exc.messages}) from exc
+
     def apply(self, data: dict[str, Any]) -> dict[str, Any]:
         """Apply value transformations to the data dictionary."""
         if not self.value_transformations:
             return data
 
         try:
-            executor = JavaScriptExecutor(data, self.value_transformations)
+            if self.engine == self.Engine.STEFICON:
```
saxix, line 81: I would suggest using Strategy Pattern here, something like:

```
class Transformer(....
    _executors = { self.Engine.STEFICON: SteficonExecutor,
                           self.Engine.STEFICON: JavaScriptExecutor}

def apply(...):
     ...
     executor = self._executors[self.engine](data, self.value_transformations)
     ...
```

and create `AbstractExecutor(ABC)` with  `apply()` signature

saxix (changes requested):

saxix (changes requested):
check comments

PR #397, opened 2026-05-28
feat: 315764 Add push action when RDP is created

roma-valor:
# Legal Boilerplate
[AB#315764](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/315764)

When you run Create and Push, the action opens a configuration page with a form.
Currently, the form contains a single field: error threshold percentage required to allow the push.
This threshold defines the maximum acceptable error rate before the push is blocked. For example, if there are 300 records and 10 errors (tickets), the push is allowed because the error rate is below the configured threshold of 90% (i.e., only up to 30 errors would be acceptable).
If the number of errors exceeds the threshold, the push will not proceed.


domdinicola (approved):

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -78,6 +94,210 @@ def create_rdp_core(job: AsyncJob) -> dict[str, Any]:
...
+            logger.info(
+                "dedup_callback_handle: rdp_id=%s skip FAILURE (%s); status=%s",
+                rdp_id,
+                reason,
+                locked.status,
+            )
+
+
+def _handle_deduplicated(rdp: Rdp, rdp_id: int, findings_count: int) -> None:
+    origin_job = AsyncJob.objects.filter(rdp_id=rdp_id).order_by("-id").first()
```
saxix: this can return None.

Let use `.latest("id")` it explain what we want to do and raises `DoesNotExist` if no records found

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -78,6 +94,210 @@ def create_rdp_core(job: AsyncJob) -> dict[str, Any]:
...
+            logger.warning(
+                "dedup_callback_handle: rdp_id=%s skip push queue; status changed to %s",
+                rdp_id,
+                locked.status,
+            )
+            return
+        locked.status = Rdp.PushStatus.PENDING
+        locked.save(update_fields=["status"])
+
+    push_job = AsyncJob.objects.create(
```
saxix: is this correct? We could have RDP locked but no asyncjob

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -78,6 +94,210 @@ def create_rdp_core(job: AsyncJob) -> dict[str, Any]:
...
+                locked.status,
+            )
+
+
+def _handle_deduplicated(rdp: Rdp, rdp_id: int, findings_count: int) -> None:
+    origin_job = AsyncJob.objects.filter(rdp_id=rdp_id).order_by("-id").first()
+    max_findings_percent: int = origin_job.config.get("max_dedup_findings_percent", 0) if origin_job else 0
+
+    total_individuals = qs_individuals_for_rdp(rdp=rdp).count()
+    findings_rate = findings_count / total_individuals * 100 if total_individuals > 0 else float(findings_count)
```
saxix: what if total_individuales is 0 ?

saxix (changes requested):

src/country_workspace/contrib/hope/push/orchestration.py
roma-valor: fixed

src/country_workspace/contrib/hope/push/orchestration.py
roma-valor: moved push_job into lock

src/country_workspace/contrib/hope/push/orchestration.py
roma-valor: resolved

src/country_workspace/models/rdp.py
```diff
@@ -21,7 +22,7 @@ class OperationAction(models.TextChoices):
     country_office = models.ForeignKey("Office", on_delete=models.CASCADE, related_name="%(class)ss")
     program = models.ForeignKey("Program", on_delete=models.CASCADE, related_name="%(class)ss")
     name = models.CharField(max_length=255, blank=True, null=True)
-    status = models.CharField(max_length=10, choices=PushStatus.choices, default=PushStatus.PENDING, blank=True)
+    status = models.CharField(max_length=15, choices=PushStatus.choices, default=PushStatus.PENDING, blank=True)
```
vitali-yanushchyk-valor: Since we’re adding another RDP status and updating status-related field, could we wrap these transitions in a helper, like?
```
# Do not rename: migrations reference this callable by dotted path.
def get_rdp_status_choices() -> list[tuple[str, str]]:
    return Rdp.PushStatus.choices
...
DEDUP_PENDING = "DEDUP_PENDING", _("Awaiting deduplication")
...
    status = models.CharField(max_length=15, choices=get_rdp_status_choices, default=PushStatus.PENDING, blank=True)
```

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -78,6 +94,220 @@ def create_rdp_core(job: AsyncJob) -> dict[str, Any]:
...
+                "dedup_callback_handle: rdp_id=%s skip FAILURE (%s); status=%s",
+                rdp_id,
+                reason,
+                locked.status,
+            )
+
+
+def _handle_deduplicated(rdp: Rdp, rdp_id: int, findings_count: int) -> None:
+    try:
+        origin_job = AsyncJob.objects.filter(rdp_id=rdp_id).latest("id")
```
vitali-yanushchyk-valor: Could this select a different job created for the same RDP?
Since the callback token already contains the originating job_id, should we use that exact job instead of relying on the latest one?

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -78,6 +94,220 @@ def create_rdp_core(job: AsyncJob) -> dict[str, Any]:
...
+            raise HopePushError({"errors": ["RDP: could not claim deduplication set."], "rdp_id": rdp_id})
+
+    job.refresh_from_db()
+
+    callback_url = _build_dedup_callback_url(rdp_id=rdp_id, job_id=job.pk)
+    processor = DedupProcessor(rdp)
+    processor.run(notification_url=callback_url)
+
+    if processor.has_errors:
+        raise HopePushError({**processor.total, "rdp_id": rdp_id})
```
vitali-yanushchyk-valor: Could we handle this failure path explicitly? At this point the RDP has already been claimed with `is_dedup_settings_locked=True`. If `processor.run()` fails or records errors, the job exits while the RDP is still `PENDING`. Since it never reaches `DEDUP_PENDING`, the callback flow will not take over, and unlike the regular deduplication flow, there is no cleanup that releases the lock.

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -78,6 +94,220 @@ def create_rdp_core(job: AsyncJob) -> dict[str, Any]:
     return {"rdp_id": rdp.id, "rdp_str": str(rdp)}
 
 
+def _build_dedup_callback_url(rdp_id: int, job_id: int) -> str:
+    """Build an absolute, signed callback URL for the dedup engine to call when dedup finishes."""
+    token = signing.dumps(
+        {"rdp_id": rdp_id, "job_id": job_id},
+        key=_DEDUP_CALLBACK_SIGN_KEY,
```
vitali-yanushchyk-valor: Should the signing key either come from environment-backed settings, or should this value be used as `salt=` instead? As written, the hard-coded `key` replaces Django’s `SECRET_KEY`, so anyone with access to the source can generate valid callback tokens. Using a separate secret from settings would make `key=` appropriate; using `salt=` would keep Django’s application secret while namespacing these tokens.

tests/contrib/hope/test_dedup_callback_view.py
```diff
@@ -0,0 +1,83 @@
+"""Tests for the dedup engine callback HTTP endpoint."""
+
+import pytest
+from django.core import signing
+from django.test import RequestFactory
+
+from country_workspace.contrib.hope.push.orchestration import (
+    _DEDUP_CALLBACK_MAX_AGE,
```
saxix: Why are we accessing "protected" attributes? I would remove the leading '_'

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -78,6 +94,220 @@ def create_rdp_core(job: AsyncJob) -> dict[str, Any]:
...
+        if rdp is None:
+            raise HopePushError({"errors": ["RDP: could not claim deduplication set."], "rdp_id": rdp_id})
+
+    job.refresh_from_db()
+
+    callback_url = _build_dedup_callback_url(rdp_id=rdp_id, job_id=job.pk)
+    processor = DedupProcessor(rdp)
+    processor.run(notification_url=callback_url)
+
+    if processor.has_errors:
```
saxix: this cal leave the set in 'PENDING' status, it this is managed using a cron job, let's add a note here to avoid confusion

saxix (changes requested):

src/country_workspace/workspaces/admin/cleaners/actions.py
```diff
@@ -248,35 +342,59 @@ def create_rdp(
     if model_admin._check_empty_queryset(request, queryset):
         return redirect(".")
     program = model_admin.get_selected_program(request)
+    show_push_option = model_admin.has_push_rdp_to_hope_permission(request) and program.biometric_deduplication_enabled
+
+    if (
```
saxix: can we simplify these nested 'if' ?

saxix (dismissed):

src/country_workspace/models/rdp.py
roma-valor: fixed

src/country_workspace/contrib/hope/push/orchestration.py
roma-valor: fixed

src/country_workspace/contrib/hope/push/orchestration.py
roma-valor: fixed

src/country_workspace/contrib/hope/push/orchestration.py
roma-valor: fixed

tests/contrib/hope/test_dedup_callback_view.py
roma-valor: fixed

src/country_workspace/contrib/hope/push/orchestration.py
roma-valor: fixed

src/country_workspace/workspaces/admin/cleaners/actions.py
roma-valor: fixed

datamik (approved):

src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -78,6 +94,235 @@ def create_rdp_core(job: AsyncJob) -> dict[str, Any]:
...
+    if not (rdp_program and rdp_program.biometric_deduplication_enabled):
+        raise HopePushError({"errors": ["RDP: push to HOPE requires biometric deduplication for this program."]})
+    return create_rdp_and_start_dedup_core(job)
+
+
+def _lock_and_fail_rdp(rdp_id: int, *, reason: str) -> None:
+    with transaction.atomic():
+        locked = lock_rdp_for_update(pk=rdp_id)
+        if locked.status == Rdp.PushStatus.DEDUP_PENDING:
+            set_rdp_push_status(rdp=locked, status=Rdp.PushStatus.FAILURE, hope_rdi_id="N/A")
```
vitali-yanushchyk-valor: Could we pass `is_dedup_settings_locked=False` here? `set_rdp_push_status()` otherwise leaves the lock unchanged, so the RDP becomes `FAILURE` while remaining dedup-locked.
Once the DE callback reports a terminal outcome, the automatic flow is finished and the RDP should be recoverable through the regular actions. With the current policy, the remaining lock blocks deduplication retry and cancellation, and a `FAILURE` RDP also prevents creating another RDP for the same program.

roma-valor: fixed

vitali-yanushchyk-valor (approved):

PR #404, opened 2026-06-04
315766: Add batch picture import functionality

arsen-vs:
- Introduced `BatchPictureImportForm` for uploading ZIP files containing images.
- Implemented methods for processing and validating uploaded images, including handling duplicates and generating previews.
- Added a new template for the import form to guide users through the upload and matching process.
- Enhanced unit tests to cover the new functionality, ensuring correct handling of image imports and matching logic.

[AB#315766](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/315766)

### DIFF
--- a/src/country_workspace/admin/batch.py
+++ b/src/country_workspace/admin/batch.py
@@ -98,6 +98,18 @@ class BatchAdmin(BaseModelAdmin):
             base = reverse("workspace:workspaces_countrybatch_changelist")
             btn.href = f"{base}?%s" % req.META["QUERY_STRING"]
 
+    @link(change_list=False, label="Import Pictures")
+    def import_pictures(self, button: LinkButton) -> None:
+        obj: Batch = button.context["original"]
+        request: HttpRequest | None = button.context.get("request")
+
+        if request and not request.user.has_perm("country_workspace.import_program_data", obj):
+            button.visible = False
+            return
+
+        workspace_change_url = obj.get_change_url(namespace="workspace")
+        button.href = workspace_change_url.replace("/change/", "/import_pictures/")
+
     @button()
     def batch_cleanup(self, request: HttpRequest, pk: str) -> HttpResponse:
         obj: Batch = self.get_object(request, pk)
--- a/src/country_workspace/admin/batch.py
+++ b/src/country_workspace/admin/batch.py
@@ -98,20 +98,6 @@ class BatchAdmin(BaseModelAdmin):
             base = reverse("workspace:workspaces_countrybatch_changelist")
             btn.href = f"{base}?%s" % req.META["QUERY_STRING"]
 
-    @link(change_list=False, label="Import Pictures")
-    def import_pictures(self, button: LinkButton) -> None:
-        obj: Batch = button.context["original"]
-        request: HttpRequest | None = button.context.get("request")
-
-        if request and not request.user.has_perm("country_workspace.import_program_data", obj):
-            button.visible = False
-            return
-
-        workspace_change_url = reverse("workspace:workspaces_countrybatch_change", args=[obj.pk])
-        if workspace_change_url.startswith("/workspaces/"):
-            workspace_change_url = f"/workspace{workspace_change_url}"
-        button.href = workspace_change_url.replace("/change/", "/import_pictures/")
-
     @button()
     def batch_cleanup(self, request: HttpRequest, pk: str) -> HttpResponse:
         obj: Batch = self.get_object(request, pk)
--- a/src/country_workspace/admin/batch.py
+++ b/src/country_workspace/admin/batch.py
@@ -107,7 +107,10 @@ class BatchAdmin(BaseModelAdmin):
             button.visible = False
             return
 
-        workspace_change_url = obj.get_change_url(namespace="workspace")
+        workspace_change_url = reverse("workspace:workspaces_countrybatch_change", args=[obj.pk])
+        workspace_root = reverse("workspace:index").rstrip("/")
+        if workspace_root and not workspace_change_url.startswith(f"{workspace_root}/"):
+            workspace_change_url = f"{workspace_root}{workspace_change_url}"
         button.href = workspace_change_url.replace("/change/", "/import_pictures/")
 
     @button()
--- a/src/country_workspace/admin/batch.py
+++ b/src/country_workspace/admin/batch.py
@@ -108,9 +108,8 @@ class BatchAdmin(BaseModelAdmin):
             return
 
         workspace_change_url = reverse("workspace:workspaces_countrybatch_change", args=[obj.pk])
-        workspace_root = reverse("workspace:index").rstrip("/")
-        if workspace_root and not workspace_change_url.startswith(f"{workspace_root}/"):
-            workspace_change_url = f"{workspace_root}{workspace_change_url}"
+        if workspace_change_url.startswith("/workspaces/"):
+            workspace_change_url = f"/workspace{workspace_change_url}"
         button.href = workspace_change_url.replace("/change/", "/import_pictures/")
 
     @button()
--- a/src/country_workspace/config/fragments/constance.py
+++ b/src/country_workspace/config/fragments/constance.py
@@ -69,6 +69,26 @@ CONSTANCE_CONFIG = {
     "DEDUP_API_URL": (DEDUP_API_URL, "Dedup Engine server address", str),
     "DEDUP_API_TOKEN": (DEDUP_API_TOKEN, "Dedup Engine API access token", "write_only_text_input"),
     "CHUNK_SIZE_FOR_VALIDATION_TASK": (500, "Number of records to process per chunk in validation tasks", int),
+    "PICTURE_IMPORT_MAX_ZIP_UPLOAD_BYTES": (
+        20 * 1024 * 1024,
+        "Maximum uploaded ZIP size in bytes for batch picture import",
+        int,
+    ),
+    "PICTURE_IMPORT_MAX_ZIP_FILE_COUNT": (
+        2000,
+        "Maximum number of files allowed in a picture import ZIP",
+        int,
+    ),
+    "PICTURE_IMPORT_MAX_ZIP_UNCOMPRESSED_BYTES": (
+        200 * 1024 * 1024,
+        "Maximum uncompressed ZIP size in bytes for picture import",
+        int,
+    ),
+    "PICTURE_IMPORT_SESSION_TTL_SECONDS": (
+        3600,
+        "How long a picture import preview remains valid before expiring",
+        int,
+    ),
     "RDP_CLEANUP_DAYS": (
         30,
         "Number of days to keep Household and Individual data after a successful RDP push. 0 means disabled.",
@@ -156,6 +176,10 @@ CONSTANCE_CONFIG_FIELDSETS = {
         "CACHE_TIMEOUT",
         "CACHE_BY_VERSION",
         "CHUNK_SIZE_FOR_VALIDATION_TASK",
+        "PICTURE_IMPORT_MAX_ZIP_UPLOAD_BYTES",
+        "PICTURE_IMPORT_MAX_ZIP_FILE_COUNT",
+        "PICTURE_IMPORT_MAX_ZIP_UNCOMPRESSED_BYTES",
+        "PICTURE_IMPORT_SESSION_TTL_SECONDS",
         "RDP_CLEANUP_DAYS",
         "RDP_CLEANUP_BATCH_SIZE",
         "NEW_USER_IS_STAFF",
--- a/src/country_workspace/config/fragments/constance.py
+++ b/src/country_workspace/config/fragments/constance.py
@@ -69,19 +69,14 @@ CONSTANCE_CONFIG = {
     "DEDUP_API_URL": (DEDUP_API_URL, "Dedup Engine server address", str),
     "DEDUP_API_TOKEN": (DEDUP_API_TOKEN, "Dedup Engine API access token", "write_only_text_input"),
     "CHUNK_SIZE_FOR_VALIDATION_TASK": (500, "Number of records to process per chunk in validation tasks", int),
-    "PICTURE_IMPORT_MAX_ZIP_UPLOAD_BYTES": (
-        20 * 1024 * 1024,
-        "Maximum uploaded ZIP size in bytes for batch picture import",
+    "PICTURE_IMPORT_MAX_ZIP_UPLOAD_MB": (
+        20,
+        "Maximum uploaded ZIP size in MB for batch picture import",
         int,
     ),
     "PICTURE_IMPORT_MAX_ZIP_FILE_COUNT": (
         2000,
         "Maximum number of files allowed in a picture import ZIP",
-        int,
-    ),
-    "PICTURE_IMPORT_SESSION_TTL_SECONDS": (
-        3600,
-        "How long a picture import preview remains valid before expiring",
         int,
     ),
     "RDP_CLEANUP_DAYS": (
@@ -171,9 +166,8 @@ CONSTANCE_CONFIG_FIELDSETS = {
         "CACHE_TIMEOUT",
         "CACHE_BY_VERSION",
         "CHUNK_SIZE_FOR_VALIDATION_TASK",
-        "PICTURE_IMPORT_MAX_ZIP_UPLOAD_BYTES",
+        "PICTURE_IMPORT_MAX_ZIP_UPLOAD_MB",
         "PICTURE_IMPORT_MAX_ZIP_FILE_COUNT",
-        "PICTURE_IMPORT_SESSION_TTL_SECONDS",
         "RDP_CLEANUP_DAYS",
         "RDP_CLEANUP_BATCH_SIZE",
         "NEW_USER_IS_STAFF",
--- a/src/country_workspace/config/fragments/constance.py
+++ b/src/country_workspace/config/fragments/constance.py
@@ -79,11 +79,6 @@ CONSTANCE_CONFIG = {
         "Maximum number of files allowed in a picture import ZIP",
         int,
     ),
-    "PICTURE_IMPORT_MAX_ZIP_UNCOMPRESSED_BYTES": (
-        200 * 1024 * 1024,
-        "Maximum uncompressed ZIP size in bytes for picture import",
-        int,
-    ),
     "PICTURE_IMPORT_SESSION_TTL_SECONDS": (
         3600,
         "How long a picture import preview remains valid before expiring",
@@ -178,7 +173,6 @@ CONSTANCE_CONFIG_FIELDSETS = {
         "CHUNK_SIZE_FOR_VALIDATION_TASK",
         "PICTURE_IMPORT_MAX_ZIP_UPLOAD_BYTES",
         "PICTURE_IMPORT_MAX_ZIP_FILE_COUNT",
-        "PICTURE_IMPORT_MAX_ZIP_UNCOMPRESSED_BYTES",
         "PICTURE_IMPORT_SESSION_TTL_SECONDS",
         "RDP_CLEANUP_DAYS",
         "RDP_CLEANUP_BATCH_SIZE",
--- a/src/country_workspace/models/batch.py
+++ b/src/country_workspace/models/batch.py
@@ -41,34 +41,24 @@ class Batch(BaseModel):
     def __str__(self) -> str:
         return self.name or f"Batch self.pk ({self.country_office})"
 
-    def get_picture_import_state(self) -> dict[str, dict[str, Any]]:
+    def get_picture_import_state(self) -> dict[str, Any]:
         state = self.picture_import_state
         if not isinstance(state, dict):
             return {}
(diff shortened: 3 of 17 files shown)


src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +163,144 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+                preview = service.build_preview(form.cleaned_data["match_field"], form.cleaned_data["zip_file"])
+                token = str(uuid.uuid4())
+                self._save_picture_import_payload(
+                    request,
+                    token,
+                    {
+                        "batch_id": obj.pk,
+                        "match_field": form.cleaned_data["match_field"],
+                        "target_field": form.cleaned_data["target_field"],
+                        **preview,
```
vitali-yanushchyk-valor: Am I right that all matching images are base64-encoded and stored in the session until confirmation?
If I’m not mistaken, `SESSION_ENGINE` is set to `django.contrib.sessions.backends.db`. Could this result in oversized rows in `django_session`?

vitali-yanushchyk-valor:
Do we need limits on archive size, file count, and total uncompressed size before reading and base64-encoding all entries?

src/country_workspace/workspaces/admin/batch/picture_import.py
```diff
@@ -0,0 +1,144 @@
...
+        updated = 0
+        for item in assignments:
+            individual = individuals.get(item["record_id"])
+            if not individual:
+                continue
+            current = dict(individual.flex_fields or {})
+            current[target_field] = item["data_uri"]
+            if current != individual.flex_fields:
+                individual.flex_fields = current
+                individual.save(update_fields=["flex_fields"])
```
vitali-yanushchyk-valor: Should importing pictures invalidate the existing validation status, for example by setting `last_checked` to `None` without triggering immediate revalidation?

src/country_workspace/workspaces/admin/batch/picture_import.py
```diff
@@ -0,0 +1,144 @@
...
+
+                content = archive.read(info)
+                data_uri = f"data:{mimetype};base64,{base64.b64encode(content).decode()}"
+                entries.append({"filename": filename, "key": key, "data_uri": data_uri})
+
+        zip_file.seek(0)
+        return entries, duplicates
+
+    def get_match_field_choices(self) -> list[tuple[str, str]]:
+        first_individual = CountryIndividual.objects.filter(batch=self.batch, removed=False).only("raw_data").first()
```
vitali-yanushchyk-valor: Are all individuals in a batch guaranteed to have the same `raw_data` keys? Otherwise, using only the first active individual may hide valid match fields that are present in other records.

src/country_workspace/workspaces/admin/batch/admin.py
arsen-vs: Solved

src/country_workspace/workspaces/admin/batch/picture_import.py
arsen-vs: Solved

src/country_workspace/workspaces/admin/batch/picture_import.py
arsen-vs: Solved

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +170,200 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+        if path:
+            zip_path = Path(path)
+            if zip_path.exists():
+                zip_path.unlink()
+
+    @staticmethod
+    def _store_uploaded_zip(uploaded: UploadedFile) -> str:
+        path: str | None = None
+        try:
+            with tempfile.NamedTemporaryFile(prefix="batch-picture-import-", suffix=".zip", delete=False) as temp_file:
```
vitali-yanushchyk-valor: Since temporary ZIPs are created with delete=False, what cleans them up if the user never confirms the import or the session expires?

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +170,200 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+                        self.message_user(
+                            request,
+                            _("Picture import session has expired. Please run the matching step again."),
+                            messages.ERROR,
+                        )
+                        response = HttpResponseRedirect(request.path)
+                    else:
+                        try:
+                            with Path(zip_temp_path).open("rb") as zip_stream:
+                                preview = service.build_preview(
```
vitali-yanushchyk-valor: On confirmation we rebuild the matching instead of applying the assignments shown to the user. Could this change the confirmed result if the batch is updated between preview and confirmation? For example, if a matched record is marked as removed or its matching raw-data key changes, the file may no longer be applied to the same record.

Could we avoid rebuilding the matches on confirmation? It may be safer to use the assignments saved during preview and only load the image data from the stored ZIP before applying them.

src/country_workspace/workspaces/admin/batch/admin.py
vitali-yanushchyk-valor: Could we use a shared Django storage instead of a local temp file here? The backend may run on multiple pods, so the preview and confirmation requests may hit different pods, and the local ZIP path may not exist on the second one.

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +170,200 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+            self._delete_temp_zip(payload.get("zip_temp_path"))
+        request.session[BATCH_PICTURE_IMPORT_SESSION_KEY] = payloads
+        request.session.modified = True
+
+    @button(
+        change_list=False,
+        permission=can_import_program_data,
+        html_attrs={"title": "Import pictures from zip and assign by matching key."},
+    )
+    def import_pictures(self, request: HttpRequest, pk: str) -> HttpResponse:  # noqa: C901, PLR0912, PLR0915
```
vitali-yanushchyk-valor, line 215: This action is getting a bit hard to follow now, since `import_pictures()` handles form setup, preview/confirmation flow, temp-file handling, session payloads, etc. Maybe we could extract some of this into small helper methods to make it easier to maintain?

src/country_workspace/workspaces/admin/batch/picture_import.py
```diff
@@ -0,0 +1,215 @@
...
+    def _get_int_config(name: str, default: int) -> int:
+        value = getattr(constance_config, name, default)
+        try:
+            parsed = int(value)
+        except (TypeError, ValueError):
+            return default
+        return parsed if parsed > 0 else default
+
+    @classmethod
+    def max_zip_upload_bytes(cls) -> int:
```
datamik: The same here. Do we need so many methods to get the value from the settings?
Also do we need to hard code default values in the class? We can just read them from the constance settings that already have the defaults set

src/country_workspace/config/fragments/constance.py
```diff
@@ -156,6 +176,10 @@
         "CACHE_TIMEOUT",
         "CACHE_BY_VERSION",
         "CHUNK_SIZE_FOR_VALIDATION_TASK",
+        "PICTURE_IMPORT_MAX_ZIP_UPLOAD_BYTES",
+        "PICTURE_IMPORT_MAX_ZIP_FILE_COUNT",
+        "PICTURE_IMPORT_MAX_ZIP_UNCOMPRESSED_BYTES",
```
datamik: Lets remove this "PICTURE_IMPORT_MAX_ZIP_UNCOMPRESSED_BYTES"

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +175,250 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+        return cleaned_payloads
+
+    @staticmethod
+    def _picture_import_session_ttl_seconds() -> int:
+        value = getattr(constance_config, "PICTURE_IMPORT_SESSION_TTL_SECONDS", 3600)
+        try:
+            parsed = int(value)
+        except (TypeError, ValueError):
+            return 3600
+        return parsed if parsed > 0 else 3600
```
datamik: This could probably be simplified a little. Since the constance value is already defined as an integer, there may not be a need to cast it again.

If we still want to enforce the > 0 constraint, it might be worth introducing a custom BoundedIntegerField for constance instead, so invalid values cannot be stored in the first place.

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +175,245 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+        now = int(time())
+        cleaned_payloads: dict[str, dict[str, Any]] = {}
+        changed = False
+        for token, payload in payloads.items():
+            if not isinstance(payload, dict):
+                changed = True
+                continue
+            created_at = payload.get("created_at")
+            if isinstance(created_at, int) and now - created_at > max_age_seconds:
+                CountryBatchAdmin._delete_stored_zip(payload.get("zip_storage_name") or payload.get("zip_temp_path"))
```
datamik: zip_storage_name, as i understand it's zip file name on storage but its a bit confusing name. 
Also what is zip_temp_path? 
Will _delete_stored_zip work ob both of them?

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +175,245 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+                changed = True
+                continue
+            cleaned_payloads[token] = payload
+        if changed:
+            request.session[BATCH_PICTURE_IMPORT_SESSION_KEY] = cleaned_payloads
+            request.session.modified = True
+        return cleaned_payloads
+
+    @staticmethod
+    def _delete_stored_zip(storage_name: str | None) -> None:
```
datamik: storage_name the same here. It's confusing. It's not storage_name its file_name probably.

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +175,245 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+            default_storage.delete(storage_name)
+
+    @staticmethod
+    def _cleanup_stale_stored_zips() -> None:
+        max_age_seconds = int(constance_config.PICTURE_IMPORT_SESSION_TTL_SECONDS)
+        now = int(time())
+        try:
+            _, files = default_storage.listdir("batch-picture-import")
+        except (OSError, NotImplementedError):
+            logger.debug("Could not list picture import storage directory", exc_info=True)
```
datamik: If we're returning immediately after this, it might be better to use logger.exception() instead of logger.debug()

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +175,245 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+            return
+        for filename in files:
+            storage_name = f"batch-picture-import/{filename}"
+            try:
+                modified_at = int(default_storage.get_modified_time(storage_name).astimezone(UTC).timestamp())
+            except (OSError, NotImplementedError):
+                logger.debug("Could not get modified time for %s", storage_name, exc_info=True)
```
datamik: The same, logger.exception()  would be more informative

src/country_workspace/workspaces/admin/batch/admin.py
datamik: O, i see you have exc_info=True, but, still, it's simpler to use exception()

src/country_workspace/config/fragments/constance.py
```diff
@@ -69,6 +69,21 @@
     "DEDUP_API_URL": (DEDUP_API_URL, "Dedup Engine server address", str),
     "DEDUP_API_TOKEN": (DEDUP_API_TOKEN, "Dedup Engine API access token", "write_only_text_input"),
     "CHUNK_SIZE_FOR_VALIDATION_TASK": (500, "Number of records to process per chunk in validation tasks", int),
+    "PICTURE_IMPORT_MAX_ZIP_UPLOAD_BYTES": (
+        20 * 1024 * 1024,
+        "Maximum uploaded ZIP size in bytes for batch picture import",
+        int,
+    ),
```
vitali-yanushchyk-valor, line 76: Do we need to expose all these values in Constance? Maybe only the upload size limit should be user-facing, preferably in MB. File count and preview TTL look more like safe internal defaults.

src/country_workspace/admin/batch.py
```diff
@@ -98,6 +98,20 @@ def view_in_workspace(self, btn: "LinkButton") -> None:
             base = reverse("workspace:workspaces_countrybatch_changelist")
             btn.href = f"{base}?%s" % req.META["QUERY_STRING"]
 
+    @link(change_list=False, label="Import Pictures")
+    def import_pictures(self, button: LinkButton) -> None:
```
vitali-yanushchyk-valor: Is there a specific reason to expose this action from common admin?  I thought that the actual flow lives in workspace admin.

src/country_workspace/workspaces/admin/batch/picture_import.py
```diff
@@ -0,0 +1,183 @@
...
+
+    @staticmethod
+    def _normalize_match_key(value: object | None) -> str:
+        if value is None:
+            return ""
+        return str(value).strip().lower()
+
+    @staticmethod
+    def _guess_image_mimetype(filename: str) -> str:
+        guessed, _ = mimetypes.guess_type(filename)
```
vitali-yanushchyk-valor, line 48: This currently infers the image type from the filename/extension. That may be fine for trusted uploads, but should we validate the actual image content as well before saving it as Base64?

src/country_workspace/workspaces/admin/batch/picture_import.py
```diff
@@ -0,0 +1,183 @@
...
+                continue
+            enriched.append({**assignment, "data_uri": data_uri})
+        return enriched
+
+    @staticmethod
+    def apply_assignments(target_field: str, assignments: list[dict[str, Any]]) -> int:
+        if not assignments:
+            return 0
+        record_ids = [item["record_id"] for item in assignments]
+        individuals = CountryIndividual.objects.filter(pk__in=record_ids).in_bulk()
```
vitali-yanushchyk-valor: This should work for the happy path because preview only creates assignments for active records in this batch. Still, since confirm uses stored preview assignments, should the update also enforce `batch` and `removed=False`?

src/country_workspace/workspaces/admin/batch/picture_import.py
```diff
@@ -0,0 +1,183 @@
...
+            individual = individuals.get(item["record_id"])
+            if not individual:
+                continue
+            current = dict(individual.flex_fields or {})
+            current[target_field] = item["data_uri"]
+            if current != individual.flex_fields:
+                individual.flex_fields = current
+                individual.last_checked = None
+                individual.errors = {}
+                individual.save(update_fields=["flex_fields", "last_checked", "errors"])
```
vitali-yanushchyk-valor: Should this update step be atomic? Right now records are updated one by one, so an exception in the middle could leave the batch partially updated. I think an all-or-nothing import would be safer here.

vitali-yanushchyk-valor:
I think there are a few remaining flow-level questions to clarify before finalizing this implementation:

1. Sessions seem fine if the picture import draft is intentionally user-private and short-lived. I just want to confirm that this is the intended behavior. If the draft/import state is expected to be visible or traceable beyond the current user session, maybe a separate DB-backed draft/import flow would fit better.

2. Preview should probably remain synchronous and lightweight, since the user needs the matching report before confirming. However, confirm currently reads image data, generates Base64, and updates records in the request. Would it make sense to schedule the actual import step as a background job?

3. Related to the preview/confirm split: preview currently uses the same ZIP extraction path as confirm, which reads image content and builds Base64 data. From the current UI/report, preview seems to need only matching metadata, so could we keep preview metadata-only, store only minimal temporary state, and rebuild assignments from the saved ZIP and current DB state during confirm/import?

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +177,263 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+                        if not batch_lock:
+                            self.message_user(
+                                request,
+                                _("Another action is currently running for this batch. Please try again later."),
+                                messages.ERROR,
+                            )
+                            response = HttpResponseRedirect(request.path)
+                        else:
+                            try:
+                                with default_storage.open(zip_file_name, "rb") as zip_stream:
```
vitali-yanushchyk-valor: Since confirm opens the stored ZIP, reads image data, generates Base64, and then updates records in the same request, would it make sense to schedule the actual import step as a background job?

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +177,263 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+                                    assignments = service.enrich_assignments_with_zip_data(
+                                        payload.get("assignments", []),
```
vitali-yanushchyk-valor: Confirm currently uses assignments generated during preview. If batch reprocessing or another update runs between preview and confirmation, the stored `record_id -> filename` mapping may become stale. Would it make sense to rebuild assignments during confirm so the final update uses the current DB state?

src/country_workspace/models/batch.py
```diff
@@ -27,6 +30,15 @@ class BatchSource(models.TextChoices):
...
     )
+    picture_import_state = models.JSONField(default=dict, blank=True)
+    picture_import_state_updated_by = models.ForeignKey(
+        User,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        related_name="picture_import_state_batches",
+    )
+    picture_import_state_updated_at = models.DateTimeField(null=True, blank=True)
```
datamik: Can we just use one json field and store all the fields state, date and user in this one json field?

arsen-vs: As for me it is bad design to use JSON fields instead of specific field variables. But in case that we already have picture_import_state as JSON, it should not change a lot. I will merge them

src/country_workspace/workspaces/admin/batch/admin.py
```diff
@@ -119,6 +176,246 @@ def has_add_permission(self, request: HttpRequest, obj: CountryBatch | None = No
...
+            logger.exception("Could not list picture import storage directory")
+            return
+        for filename in files:
+            file_name = f"batch-picture-import/{filename}"
+            try:
+                modified_at = int(default_storage.get_modified_time(file_name).astimezone(UTC).timestamp())
+            except (OSError, NotImplementedError):
+                logger.exception("Could not get modified time for %s", file_name)
+                continue
+            if now - modified_at > PICTURE_IMPORT_SESSION_TTL_SECONDS:
```
vitali-yanushchyk-valor: Now that confirmation schedules a background job, could the TTL-based cleanup delete a ZIP before the queued import job starts?

src/country_workspace/workspaces/templates/workspace/admin_extra_buttons/import_pictures_form.html
```diff
@@ -0,0 +1,121 @@
...
+                        <form method="post" class="mt-6">
+                            {% csrf_token %}
+                            <input type="hidden" name="token" value="{{ token }}" />
+                            <input type="hidden" name="confirm" value="yes" />
+                            <input type="submit" class="button default" value="{% translate 'Confirm import' %}" />
+                            <a class="button" href="{{ request.path }}">{% translate "Start over" %}</a>
+                        </form>
+                    {% else %}
+                        <p class="mt-2">
+                            {% translate "Step 3: confirmation." %}
```
vitali-yanushchyk-valor: Very minor: since confirmation now schedules a background job and redirects, maybe this should say that the import was scheduled rather than completed?

src/country_workspace/workspaces/admin/batch/picture_import.py
```diff
@@ -0,0 +1,236 @@
...
+            "matched_files_count": len(assignments),
+            "duplicate_zip_keys": sorted(duplicate_zip_keys),
+            "ambiguous_record_keys": sorted(ambiguous_record_keys),
+            "unmatched_filenames": sorted(unmatched_filenames),
+            "assignments": assignments,
+        }
+
+    @classmethod
+    def enrich_assignments_with_zip_data(
+        cls, assignments: list[dict[str, Any]], zip_file: UploadedFile
```
vitali-yanushchyk-valor: Minor: is this helper still used?

datamik (approved):

vitali-yanushchyk-valor (approved):

PR #405, opened 2026-06-05
feat: Add normalization for invalid multiselect values in Validable model

arsen-vs:
* Implemented a static method to normalize invalid values for fields that allow multiple selections.
* Updated the validate_with_checker method to utilize the new normalization method, ensuring invalid multiselect values are preserved as lists.
* Added a test to verify that invalid multiselect values are correctly handled and stored as lists.

[AB#319150](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/319150)

### DIFF
--- a/src/country_workspace/models/base.py
+++ b/src/country_workspace/models/base.py
@@ -123,9 +123,10 @@ class Validable(Cachable, models.Model):
         raise NotImplementedError
 
     @staticmethod
-    def _normalize_invalid_value_for_field(field: Any, value: Any) -> Any:
+    def _coerce_multiselect_value(field: Any, value: Any) -> Any:
         if not getattr(getattr(field, "widget", None), "allow_multiple_selected", False):
             return value
+
         if value is None:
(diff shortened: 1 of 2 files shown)


roma-valor (approved):

vitali-yanushchyk-valor (approved):

datamik (changes requested):
It's not clear for me what are we trying to fix and what should be the expected results

datamik:
@arsen-vs 

We already have multiple-choice fields that are working properly. There is a CustomMultipleChoiceField that handles comma- or space-separated string input.

You only need to register this field by adding:

field_registry.register(CustomMultipleChoiceField)

to src/country_workspace/contrib/hope/apps.py.

After that, CustomMultipleChoiceField will appear as a field type in the field definitions. Users can then create a new field definition using this type and finally create or update a flex field to use the newly created field definition.

PR #406, opened 2026-06-07
Add INFORM / ONA Registration Data Import connector

kaltaher:
[AB#323071](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/323071)

## Summary

This PR introduces an INFORM / ONA Registration Data Import connector for Country Workspace.

The implementation follows the existing Country Workspace import architecture and avoids introducing a separate system. It reuses the current import processor, post-processing, validation, batch, and admin import patterns used by RDI/Kobo/Aurora flows.

## What changed

* Added `country_workspace.contrib.ona`
* Added ONA API client with token authentication, metadata retrieval, submission retrieval, pagination, and error handling
* Added ONA JSON-to-record transformer
* Added ONA import processing workflow
* Added `Batch.BatchSource.ONA`
* Added migration `0055_add_ona_batch_source`
* Added ONA support in batch reprocessing
* Added ONA import admin form and import tab
* Added Constance/env settings:

  * `ONA_API_TOKEN`
  * `ONA_API_URL`
  * `ONA_HH_FIELDS_TO_IGNORE`
  * `ONA_IND_FIELDS_TO_IGNORE`
* Added documentation under `docs/src/import_data/ona.md`
* Added tests for ONA client, transformer, and import processing

## Security

ONA credentials are read from Constance/env settings.

The ONA token is not stored in:

* the import form
* `AsyncJob.config`
* test fixtures using real credentials

## Test result

Local ONA tests passed:

```text
25 passed
```

Test command used:

```bash
DATABASE_URL=sqlite:///local.sqlite3 \
SECRET_KEY=local \
ALLOWED_HOSTS=localhost,127.0.0.1 \
CACHE_URL=redis://localhost:6379/15 \
CELERY_BROKER_URL=memory:// \
CSRF_COOKIE_SECURE=False \
CSRF_TRUSTED_ORIGINS=http://localhost \
MEDIA_ROOT=./local-media \
STATIC_ROOT=./local-static \
ADMIN_EMAIL=admin@example.com \
ADMIN_PASSWORD=admin \
ONA_API_TOKEN=dummy-token \
ONA_API_URL=https://api.ona.io \
PYTHONPATH=src \
uv run python -m pytest tests/contrib/ona -q -o addopts="" --confcutdir=tests/contrib/ona
```

## Safety notes

* No real ONA API call was made during development/testing
* No real INFORM token was used
* No production database was used
* No HOPE push is triggered by this connector
* The connector only imports into the Country Workspace batch/household/individual flow


kaltaher:
Hi team,

I opened this as a Draft PR for review and guidance before it moves further.

This is an initial proposal for an INFORM / ONA Registration Data Import connector in Country Workspace, following the existing Kobo/Aurora/RDI import architecture rather than introducing a separate system.

The connector currently:

* reuses the existing import processing flow
* uses `build_import_processor`
* uses `run_batch_postprocessing`
* supports validation jobs
* adds `BatchSource.ONA`
* adds an ONA import form and admin tab
* reads ONA credentials from Constance/env settings
* does not store the ONA token in `AsyncJob.config`
* does not trigger any HOPE push directly

Local ONA tests passed: `25 passed`.

Safety notes:

* No real ONA API call was made during development/testing.
* No real INFORM token was used.
* No production database was used.
* No HOPE push was triggered.

Could you please review the approach and advise whether this connector should remain in Country Workspace, or if you prefer a different location/design before this moves beyond draft?

kaltaher:
Small follow-up update: I updated the ONA import documentation to reflect the INFORM default endpoint configuration.

The default `ONA_API_URL` is now documented as:

`https://data.inform.unicef.org`

This matches the actual INFORM endpoint pattern:

`https://data.inform.unicef.org/api/v1/data/{form_id}`

The endpoint remains configurable through Constance/env via `ONA_API_URL`.

kaltaher:
The failed checks do not appear related to the implementation.

- `Adds labels / labels pull requests` fails at `Add Migration label` because the workflow token only has read permissions (`contents: read`, `pullRequests: read`) and cannot add labels.
- `Documentation / Generate` succeeds, but `Documentation / deploy` fails after generation, likely due to GitHub Pages deployment permissions/environment restrictions.

Could a maintainer please approve/rerun the workflows or adjust the workflow permissions?

src/country_workspace/contrib/ona/forms.py
```diff
@@ -0,0 +1,37 @@
+from typing import Any
+
+from django import forms
+
+from country_workspace.models import Program
+from country_workspace.workspaces.admin.forms import BaseImportForm
+
+
+class ImportOnaForm(BaseImportForm):
+    form_id = forms.CharField(
```
saxix: this is a bit dangerous, allowing importing data from other business areas/programmes. We need to find a "protocol" to strictly li link ONA Form to CW Offices/Programmes

saxix (changes requested):
please rebase

kaltaher:
Thanks, understood.

I agree that the current free-text `form_id` field is not the right final approach.

The intended design should be closer to the existing Kobo import flow: the user should not manually type an arbitrary source/form ID. Instead, the ONA/INFORM import should expose only approved forms that are allowed for the current Country Workspace office/programme.

I will update the draft so:

- `form_id` is no longer an unrestricted free-text field;
- ONA/INFORM forms are shown as controlled choices, similar to Kobo projects;
- the selected form is validated server-side before the import job is created;
- the selected form must be explicitly allowed for the current CW office/programme;
- the import fails closed if no approved mapping exists.

This should address the concern about importing data from other business areas/programmes.

kaltaher:
Updated, thanks.

I rebased the branch on the latest `develop` and pushed the updated history.

I also changed the ONA/INFORM import form so `form_id` is no longer unrestricted free text. It now follows the same principle as the Kobo import flow: forms must be explicitly approved/configured and are shown as controlled choices for the current Country Workspace office/programme.

Main changes:

* added `ONA_APPROVED_FORMS` configuration;
* changed `form_id` from `CharField` to `ChoiceField`;
* added server-side validation to ensure the selected ONA/INFORM form is approved for the current office/programme;
* made the flow fail closed if no approved mapping exists;
* added tests for allowed, blocked, and invalid mappings;
* updated documentation.

Local ONA tests pass: `33 passed`.

src/country_workspace/contrib/ona/import_processing.py
```diff
@@ -0,0 +1,287 @@
...
+    )
+
+    return Individual.objects.create(
+        batch_id=batch.pk,
+        name="",
+        originating_id=originating_id,
+        household=household,
+        flex_fields=individual_row_processor(row),
+        raw_data={
+            "fields": dict(row),
```
roma-valor: when processor calls it, it expects flat row, it does not unwrap 'fields', so the data will be corrupted

src/country_workspace/workspaces/admin/_import_data.py
```diff
@@ -256,3 +267,54 @@ def import_kobo(self, request: HttpRequest, program: "CountryProgram") -> "Impor
...
 
         return form
+
+    def import_ona(self, request: HttpRequest, program: "CountryProgram") -> "ImportOnaForm | None":
+        form = ImportOnaForm(request.POST, prefix="ona", program=program)
+        if form.is_valid():
+            config: OnaConfig = {
+                "batch_name": form.cleaned_data["batch_name"] or batch_name_default(),
+                "validate_after_import": bool(form.cleaned_data.get("validate_after_import")),
+                "fail_if_alien": bool(form.cleaned_data.get("fail_if_alien")),
```
roma-valor: Do we define this field anywhere?

roma-valor (changes requested):

roma-valor:
Few additional comments:
- Can we use SyncLog/last_id to skip already-imported records like Kobo/Aurora uses?
- Can we add newlines where it's missed?

kaltaher:
Thanks for the review.

Understood on both points. I’ll update the ONA import so `raw_data` remains a flat row instead of nesting the row under `fields`, and I’ll remove the undefined `fail_if_alien` config entry from the ONA import flow unless it is explicitly added to the form later.

I’ll update the tests and push a follow-up.

kaltaher:
Updated, thanks.

I fixed both review points:

* ONA `raw_data` now remains a flat row instead of nesting the row under `fields`;
* the original ONA submission is preserved under `_ona_source_submission`;
* removed the undefined `fail_if_alien` config entry from the ONA import flow;
* added tests for the flat `raw_data` behavior.

Local ONA tests pass: `35 passed`.

kaltaher:
Updated again, thanks.

I rebased the branch on the latest `upstream/develop` and resolved the conflict in `BatchSource` / `picture_import_state`.

The earlier review fixes are still included:

* ONA `raw_data` remains flat;
* original ONA submission is preserved under `_ona_source_submission`;
* removed the undefined `fail_if_alien` config entry from the ONA import flow;
* added tests for flat `raw_data` behavior.

Local ONA tests pass after rebase: `35 passed`.

src/country_workspace/contrib/ona/import_processing.py
kaltaher: Fixed  `raw_data` now remains flat, and the original ONA submission is preserved under `_ona_source_submission`.

src/country_workspace/workspaces/admin/_import_data.py
kaltaher: Fixed removed `fail_if_alien` from the ONA import config because it is not defined on the ONA form.

src/country_workspace/contrib/ona/import_processing.py
```diff
@@ -0,0 +1,287 @@
...
+    client = OnaClient(
+        base_url=constance_config.ONA_API_URL,
+        token=constance_config.ONA_API_TOKEN,
+    )
+
+    total_people = 0
+    total_households = 0
+
+    try:
+        for submission in client.iter_submissions(config["form_id"]):
```
vitali-yanushchyk-valor: This partial-import flow has no persisted cursor. If the ONA import fails, a retry starts from the beginning and can reprocess submissions that were already saved. Since we already use `SyncLog` for resumable Aurora and Kobo imports, could we apply the same concept here and store the last successfully imported ONA submission ID in `SyncLog`?

Btw, Roma already mentioned using `SyncLog` to skip already-imported records above.

kaltaher: Thanks, agreed. This is not implemented yet for the ONA flow.
I will review the existing Aurora/Kobo `SyncLog` usage and update the ONA import so it can persist the last successfully imported ONA submission ID and skip already imported submissions on retry.

kaltaher: @vitali-yanushchyk-valor 

I added `SyncLog` support to the ONA import flow, so it now persists the last successfully imported ONA submission cursor and skips already-imported submissions on retry.

Implementation summary:

-  added an ONA SyncLog name based on the form ID;
-  read the existing `SyncLog.last_id` for the current programme before import;
- skip submissions up to the stored cursor;
- update `SyncLog.last_id` after successful progress, including partial progress if a later submission fails;
-  added tests for skipping already-imported submissions and persisting the last successful cursor on error.

Local ONA tests pass: `39 passed`.

src/country_workspace/contrib/ona/import_processing.py
```diff
@@ -0,0 +1,338 @@
...
+            "INFORM/ONA import failed",
+            extra={
+                "batch_id": batch.pk,
+                "form_id": config.get("form_id"),
+                "submission_id": current_submission_id,
+                "last_successful_submission_id": last_successful_id,
+            },
+        )
+        raise
+    finally:
```
vitali-yanushchyk-valor, line 181: `SyncLog` now prevents reprocessing already imported submissions, but the ONA retry flow still creates a new batch every time.
Example failure scenario: the job creates Batch_1, imports submissions 1–100, then fails during post-processing or validation scheduling before marking the batch as `COMPLETE`. This `finally` block can still advance `last_id` to 100, while Batch_1 can remain `LOADING` with those records attached. The next retry creates Batch_2, sees `last_id = 100`, and skips submissions 1–100, so they remain in the original `LOADING` batch and are never part of the successful retry batch.
Could we align this with the Kobo resumable flow: reuse/lock the existing batch for retries or rescheduled runs, and only advance the cursor in a way that keeps the saved records and batch lifecycle consistent?

src/country_workspace/contrib/ona/import_processing.py
```diff
@@ -0,0 +1,338 @@
...
+    return str(value)
+
+
+def get_ona_submission_cursor_id(submission: Mapping[str, Any]) -> str:
+    value = (
+        submission.get("_id")
+        or submission.get("id")
+        or submission.get("_uuid")
+        or submission.get("uuid")
+    )
```
vitali-yanushchyk-valor: Can we rely on ONA/INFORM submissions always having a numeric, monotonic `_id`/`id` that is safe to use as a resume cursor? The current fallback to `_uuid`/`uuid` makes the cursor semantics unsafe: UUIDs are stable identities, but they are not ordered cursors, so lexicographic comparison can incorrectly skip or reprocess submissions.
Kobo avoids this by treating the cursor as numeric only (`last_id` is read as `int`, submissions are requested with `min_id`, and `submission.id` is stored back into `SyncLog`).
If ONA can return submissions without a numeric `_id`/`id`, we should define a different ordered cursor strategy instead of comparing UUIDs as strings.

kaltaher: Updated, thanks.

I aligned the ONA resumable flow more closely with the Kobo approach:

- retries now reuse and lock the existing job.batch instead of creating a new batch every time;
- a new batch is created only when the job has no batch yet;
- the resume cursor now requires a numeric _id/id;
- UUID values are no longer used for cursor ordering, and remain only for stable originating IDs;
- tests were updated for batch reuse, numeric cursor handling, skipping already-imported submissions, and persisting the last successful cursor.

Local ONA tests pass: 41 passed.

domdinicola:
@kaltaher can we please fix conflicts?

src/country_workspace/contrib/ona/client.py
```diff
@@ -0,0 +1,94 @@
...
+                start=start,
+                limit=self.page_size,
+            )
+
+            if not submissions:
+                break
+
+            yield from submissions
+
+            if len(submissions) < self.page_size:
```
vitali-yanushchyk-valor: A short page does not guarantee that there are no more submissions, so the batch may be marked complete too early. Could we use here a more reliable pagination approach and stop only after an empty response?

src/country_workspace/contrib/ona/client.py
```diff
@@ -0,0 +1,94 @@
...
+    def get_submissions_page(
+        self,
+        *,
+        form_id: str | int,
+        start: int,
+        limit: int | None = None,
+    ) -> list[dict[str, Any]]:
+        data = self.get(
+            f"/api/v1/data/{form_id}",
+            params={
```
vitali-yanushchyk-valor: The saved watermark is only applied after all submissions are fetched, and their order is not guaranteed. ONA supports filtering and sorting, so could we request only submissions with `_id > last_id`, sorted by `_id`? This would avoid reloading old submissions and prevent records from being skipped or imported again.

src/country_workspace/contrib/ona/import_processing.py
```diff
@@ -0,0 +1,336 @@
...
+                config=config,
+            )
+
+            total_people += imported.people
+            total_households += imported.households
+            last_successful_id = current_submission_id
+
+        job.ensure_not_cancelled(refresh=True)
+
+        run_batch_postprocessing(
```
vitali-yanushchyk-valor, line 137: ONA links the imported individuals to their household, but I do not see any logic that resolves and stores the Head of Household, Primary Collector, or Alternate Collector references. How are these references expected to be determined from the ONA data?

src/country_workspace/contrib/ona/import_processing.py
```diff
@@ -0,0 +1,336 @@
...
+    )
+
+    return Household.objects.create(
+        batch_id=batch.pk,
+        name="",
+        originating_id=originating_id,
+        flex_fields=household_row_processor(row),
+        raw_data={
+            **dict(row),
+            "_ona_source_submission": dict(raw_submission),
```
vitali-yanushchyk-valor: `raw_data` should preserve the original source data for that specific household or individual record, before any field mapping or import processing, because it is used as the input for batch reprocessing. Here it contains the already mapped `row` and embeds the full submission under `_ona_source_submission`.

src/country_workspace/contrib/ona/client.py
```diff
@@ -0,0 +1,94 @@
...
+    def headers(self) -> dict[str, str]:
+        return {
+            "Authorization": f"Token {self.token}",
+            "Accept": "application/json",
+        }
+
+    def get(self, path: str, params: dict[str, Any] | None = None) -> Any:
+        url = f"{self.base_url}/{path.lstrip('/')}"
+
+        response = requests.get(
```
vitali-yanushchyk-valor: Any temporary ONA API failure currently stops the whole import without retrying the request. To make the import more resilient to short-lived network or API issues, could we use a configured `Session` that retries failed GET requests with backoff, as the Kobo client does?

src/country_workspace/contrib/ona/transformers.py
```diff
@@ -0,0 +1,122 @@
...
+    For non-master-detail:
+        returns one individual record.
+
+    For master-detail:
+        returns one household record plus list of individual records.
+    """
+    household_field_mapping = household_field_mapping or {}
+    individual_field_mapping = individual_field_mapping or {}
+
+    source_metadata = {
```
vitali-yanushchyk-valor: The `source_submission_*` metadata and `source_individual_index` are merged into the beneficiary data after field mapping. They can therefore be stored in `flex_fields` as alien fields, cause validation errors, or overwrite a mapped value with the same name. Could we keep this metadata in the raw source data and exclude it from beneficiary field processing?

src/country_workspace/contrib/ona/import_processing.py
```diff
@@ -0,0 +1,336 @@
...
+        if config.get("validate_after_import"):
+            create_validation_jobs(
+                description=f"Validate records for batch {batch.pk}",
+                owner=job.owner,
+                program=job.program,
+                queryset=_validation_queryset(batch, config),
+            )
+
+        batch.status = Batch.BatchStatus.COMPLETE
+        batch.save(update_fields=["status"])
```
vitali-yanushchyk-valor, line 155: Bitcaster integration is a new feature, and the ONA import flow does not yet send `data_imported_signal`. Could we emit it after a successful ONA import, as Kobo and RDI do, so completed ONA batches are also reported to Bitcaster?

kaltaher:
Addressed the review feedback in d51f114e:

- Added Kobo-style HTTP retries/backoff, preserving ONA authentication and error handling.
- Added explicit tests for `_id > last_id` filtering and ascending `_id` sorting.
- Updated pagination, SyncLog resume, batch reuse, raw-data separation, household/collector references, batch validation, and the data-imported signal.
- Resolved the migration conflict introduced by syncing with upstream.

Local validation: 54 ONA tests and 156 related Kobo/shared-import tests passed. Lint, formatting, whitespace, and the application migration-state check passed.

src/country_workspace/contrib/ona/import_processing.py
kaltaher: Updated in d51f114e. Retries now lock and reuse the existing `job.batch`; a new batch is created only when the job has none. Previously saved records therefore remain in the batch used for post-processing and validation on retry. The saved SyncLog cursor skips imported submissions, and the reused batch is marked complete after successful processing.

src/country_workspace/contrib/ona/client.py
kaltaher: Fixed in d51f114e. Pagination now continues until ONA returns an empty response. The pagination test covers a short, non-empty page followed by another request, verifying that the short page does not terminate the import.

src/country_workspace/contrib/ona/client.py
kaltaher: Fixed in d51f114e. The saved cursor is passed to the client, which requests `{"_id": {"$gt": last_id}}` with ascending `{"_id": 1}` sorting. Tests explicitly decode and verify both parameters on every pagination request, including a zero cursor and requests without a saved cursor.

src/country_workspace/contrib/ona/import_processing.py
kaltaher: Implemented in d51f114e. After individual creation, the flow uses the processed `relationship` and `role` values: `HEAD` identifies the household head, and `PRIMARY`/`ALTERNATE` identify collectors. Their IDs are saved in the corresponding household reference fields. These values need to be supplied through the configured mapping/import processing. `NON_BENEFICIARY` records use the shared external-collector helper. Added tests for all three references and external-collector routing.

src/country_workspace/contrib/ona/import_processing.py
kaltaher: Fixed in d51f114e. Transformation now separates mapped `fields` from original `raw_data`. Individual records retain their original source rows, and household raw data excludes the configured top-level individual repeat. Record creation stores this raw data directly, without adding the `_ona_source_submission` wrapper. Tests verify the separation between raw source values and processed fields.

src/country_workspace/contrib/ona/client.py
kaltaher: Implemented in `d51f114e` using a configured `Session`, `Retry`, and `HTTPAdapter`, following the Kobo pattern. 
GET requests now retry up to three times with exponential backoff for `429`, `502`, `503`, and `504` responses, while respecting the `Retry-After` header. 

ONA token authentication and ONA-specific error handling are preserved.
I also added tests covering successful recovery, exhausted retries, backoff behavior, and `Retry-After` handling.

src/country_workspace/contrib/ona/transformers.py
kaltaher: Fixed in d51f114e. The transformer no longer adds source_submission_* or source_individual_index to the mapped beneficiary fields. 
The original source values are kept separately in each record’s raw_data, so they are excluded from flex_fields processing and cannot create alien fields or overwrite mapped values. The updated tests verify this separation

src/country_workspace/contrib/ona/import_processing.py
kaltaher: Implemented in `d51f114e`. After successful processing and marking the batch complete, the ONA flow now emits `data_imported_signal` with the programme ID, batch ID, imported record count, and ONA source, following the Kobo/RDI pattern. Added a test verifying the signal payload. The signal is mocked in local tests, so live Bitcaster delivery has not been tested.

src/country_workspace/migrations/0063_merge_ona_batch_source.py
```diff
@@ -0,0 +1,22 @@
+from django.db import migrations, models
```
roma-valor: Conflicting migrations detected: (0063_alter_rdp_country_office_and_more, 0063_merge_ona_batch_source in country_workspace)

src/country_workspace/contrib/ona/client.py
```diff
@@ -0,0 +1,112 @@
...
+                start=start,
+                limit=self.page_size,
+                last_id=last_id,
+            )
+
+            if not submissions:
+                break
+
+            yield from submissions
+            start += self.page_size
```
roma-valor: Let's add number of submissions here instead of self.page_size, theoretically we might have less than self.page_size submissions in a page.

src/country_workspace/contrib/ona/forms.py
```diff
@@ -0,0 +1,150 @@
...
+    return {_normalise(value) for value in values if value is not None}
+
+
+def get_approved_ona_forms() -> dict[str, dict[str, Any]]:
+    raw_config = getattr(constance_config, "ONA_APPROVED_FORMS", "") or "{}"
+
+    try:
+        parsed_config = json.loads(raw_config)
+    except json.JSONDecodeError as exc:
+        raise forms.ValidationError("ONA_APPROVED_FORMS must be valid JSON.") from exc
```
roma-valor: A ValidationError raised outside form validation isn't caught by Django, so it raises 500. Prefer logging and return {} (the guard at 140-142 already handles empty choices)

roma-valor (changes requested):

src/country_workspace/migrations/0064_merge_ona_batch_source.py
```diff
@@ -0,0 +1,22 @@
+from django.db import migrations, models
```
kaltaher: Fixed in 8409d998.
I synced the branch with the latest develop, renamed the ONA merge migration to 0064_merge_ona_batch_source.py, and updated its dependency to 0063_alter_rdp_country_office_and_more.

src/country_workspace/contrib/ona/client.py
kaltaher: Fixed in 8409d998. 
The pagination offset now advances by len(submissions) instead of self.page_size, so short pages are handled correctly without skipping records. I also updated the pagination test to cover this case

src/country_workspace/contrib/ona/forms.py
kaltaher: Fixed in 8409d998. 
Invalid or malformed ONA_APPROVED_FORMS configuration is now logged and returns {} instead of raising ValidationError outside form validation. The existing empty-choice guard then handles the case gracefully. 
I also added tests for malformed JSON and non-object JSON configuration

PR #407, opened 2026-06-08
feat: 319440 Add data decryption flow

roma-valor:
# Legal Boilerplate
[AB#319440](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/319440)

Country Workspace now requests encrypted responses from Aurora's records endpoint when AURORA_PAYLOAD_ENCRYPTION_KEY is configured. A new AuroraPayloadDecryptor class handles decryption, and AuroraClient.get() is introduced for the records fetch path. Failed decryption is logged at ERROR and raises RemoteError. If the key is not set, the client falls back to plain JSON automatically.


src/country_workspace/contrib/aurora/crypto.py
```diff
@@ -0,0 +1,61 @@
...
+
+        Raises:
+            RemoteError: If decryption fails due to an invalid/expired key,
+                         or if the response structure is malformed.
+
+        """
+        try:
+            outer = json.loads(response_text)
+            token: str = outer["payload"]
+            plaintext: bytes = self._fernet.decrypt(token.encode("utf-8"))
```
vitali-yanushchyk-valor, line 64: Should malformed envelopes like `{"payload": null}` be converted to `RemoteError` instead of raising raw CW-side errors such as `AttributeError`?

src/country_workspace/contrib/aurora/client.py
```diff
@@ -49,24 +55,48 @@ def _get_url(self, path: str) -> str:
...
+    ) -> dict[str, Any]:
+        """Fetch a single paginated page, decrypt if the response signals it."""
+        try:
+            ret = self.session.get(url, params=params, timeout=TIMEOUTS, headers=extra_headers)  # nosec
+            ret.raise_for_status()
+        except requests.RequestException as e:
+            raise RemoteError(f"Remote Error fetching {url}: {e}") from e
+
+        content_type = ret.headers.get("Content-Type", "")
+        if self._decryptor and ENCRYPTED_CONTENT_TYPE in content_type:
```
vitali-yanushchyk-valor, line 97: The PR description says that plain JSON is used when the key is not set. Should plain JSON also be accepted when `AURORA_PAYLOAD_ENCRYPTION_KEY` is configured? Right now non-encrypted JSON responses still fall through to `ret.json()`.

src/country_workspace/config/__init__.py
```diff
@@ -54,6 +54,13 @@ class Group(Enum):
     ),
     "AURORA_API_TOKEN": (str, "", "", False, "Aurora API token"),
     "AURORA_API_URL": (str, "", "", False, "Aurora API url"),
+    "AURORA_PAYLOAD_ENCRYPTION_KEY": (
```
saxix, line 57: Encryption Key is per Registration, not global.

saxix (changes requested):

src/country_workspace/config/__init__.py
roma-valor: @saxix 
Why is this necessary?
Then we have to save the encryption key in DB, which is less secure for in my opinion.
If it's the requirement, I'll change it.

src/country_workspace/contrib/aurora/crypto.py
roma-valor: Fixed

src/country_workspace/contrib/aurora/client.py
roma-valor: This flow is intentional, why do we need such behavior?

src/country_workspace/config/__init__.py
saxix: the encryption is "per registration", each registration can create its pair of public/public keys

saxix (changes requested):

src/country_workspace/config/__init__.py
roma-valor: @saxix 
If we do like that, then we need we need to save public/private key in db, for each registration.
Does this make sense to have private key in DB? What is the use case of it?

roma-valor: @saxix

saxix: when a registration is created, user can opt for pub/priv key encryption. This is only to secure data in Aurora that is more exposed. It is safe to have key in the db, because if you have access to the db there is no need for the key at all.

PR #411, opened 2026-06-23
320213: Refactor flex fields handling and introduce migration for fil…

arsen-vs:
…e field separation

- Updated the DedupProcessor to yield images from RDP individuals using a new method for fetching flex values.
- Enhanced the create_individuals function in Kobo sync to separate text and file fields, improving data handling.
- Added a migration to split file fields from flex fields, ensuring better data organization and integrity.
- Introduced utility functions for encoding and decoding flex files, and for merging and splitting flex payloads.
- Updated various models and admin components to utilize the new flex fields structure, enhancing overall functionality.

[AB#320213](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/320213)
[AB#318973](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/318973)


datamik (changes requested):
I have a few comments about the implementation that I think we should discuss before merging. The PR mostly reorganizes where the data is stored without fundamentally changing how it's consumed. The separation exists in the database, but by the time the data reaches the browser or the validation logic, it's merged back together again

1. Consider JSONField instead of BinaryField for flex_files

Since the data in flex_files is JSON ({field_name: base64_string}), I think if we change it to a JSONField we'd avoid the entire encode_flex_files_blob / decode_flex_files_blob layer completely — no manual json.loads, no memoryview/bytearray handling, no UnicodeDecodeError catching. Just self.flex_files.get(field_name) like we already do with flex_fields. Also, FLEX_FILES_PREFIX = 8192 means we only hash the first 8KB for checksums — a single photo can be 500KB+, so different records could produce the same checksum.

2. Implicit normalization in save() is fragile

normalize_flex_storage() in save() creates a hidden contract: "put file data into flex_fields, save will sort it out." Both validate_with_checker() and the admin form save rely on this. Any code that reads flex_fields between assignment and save will see mixed data, defeating the purpose of separation.

3. _checker_file_fields() silently swallows errors

It catches bare Exception and returns an empty set. If the checker is misconfigured, normalize_flex_storage silently skips splitting — file data stays in flex_fields with no log, no error, no indication anything went wrong.

4. Individual record view still merges everything

The spec says this should eliminate browser crashes. But when editing a single record, merge_flex_payload combines everything back together, including all Base64 images. The defer on list views helps, but the original problem — opening a record with large images — isn't solved.

5. Inconsistent application across the codebase

Kobo import splits flex_files for individuals but not households.
Many code paths still read flex_fields directly (name_parser, concatenate, mass_update, regex, collision, history_view) — these will never see file data after migration. There's no enforced rule about when to use get_flex_value() vs direct access.

6. Should the file/text field distinction live in the flex_fields library?

The detection of "which fields are file fields" comes from DataChecker/FieldDefinition. Would it make more sense to push this into hope_flex_fields so any consumer gets consistent text/file separation?

7. Migration (minor)

Migration does one-by-one UPDATEs — no bulk_update, no batching, no progress logging.

8. normalize_flex_storage can silently drop existing file data

The line file_values = {k: v for k, v in existing_file_values.items() if k not in file_fields} discards all current file field values from flex_files, expecting them to be re-provided in flex_fields. If the data is already properly split (file data in flex_files, text data in flex_fields), a save that only touches text fields will wipe out the file data. It only works today because callers like validate_with_checker() happen to merge everything back into flex_fields before saving.

arsen-vs:
> I have a few comments about the implementation that I think we should discuss before merging. The PR mostly reorganizes where the data is stored without fundamentally changing how it's consumed. The separation exists in the database, but by the time the data reaches the browser or the validation logic, it's merged back together again
> 
> 1. Consider JSONField instead of BinaryField for flex_files
> 
> Since the data in flex_files is JSON ({field_name: base64_string}), I think if we change it to a JSONField we'd avoid the entire encode_flex_files_blob / decode_flex_files_blob layer completely — no manual json.loads, no memoryview/bytearray handling, no UnicodeDecodeError catching. Just self.flex_files.get(field_name) like we already do with flex_fields. Also, FLEX_FILES_PREFIX = 8192 means we only hash the first 8KB for checksums — a single photo can be 500KB+, so different records could produce the same checksum.
> 
> 2. Implicit normalization in save() is fragile
> 
> normalize_flex_storage() in save() creates a hidden contract: "put file data into flex_fields, save will sort it out." Both validate_with_checker() and the admin form save rely on this. Any code that reads flex_fields between assignment and save will see mixed data, defeating the purpose of separation.
> 
> 3. _checker_file_fields() silently swallows errors
> 
> It catches bare Exception and returns an empty set. If the checker is misconfigured, normalize_flex_storage silently skips splitting — file data stays in flex_fields with no log, no error, no indication anything went wrong.
> 
> 4. Individual record view still merges everything
> 
> The spec says this should eliminate browser crashes. But when editing a single record, merge_flex_payload combines everything back together, including all Base64 images. The defer on list views helps, but the original problem — opening a record with large images — isn't solved.
> 
> 5. Inconsistent application across the codebase
> 
> Kobo import splits flex_files for individuals but not households. Many code paths still read flex_fields directly (name_parser, concatenate, mass_update, regex, collision, history_view) — these will never see file data after migration. There's no enforced rule about when to use get_flex_value() vs direct access.
> 
> 6. Should the file/text field distinction live in the flex_fields library?
> 
> The detection of "which fields are file fields" comes from DataChecker/FieldDefinition. Would it make more sense to push this into hope_flex_fields so any consumer gets consistent text/file separation?
> 
> 7. Migration (minor)
> 
> Migration does one-by-one UPDATEs — no bulk_update, no batching, no progress logging.
> 
> 8. normalize_flex_storage can silently drop existing file data
> 
> The line file_values = {k: v for k, v in existing_file_values.items() if k not in file_fields} discards all current file field values from flex_files, expecting them to be re-provided in flex_fields. If the data is already properly split (file data in flex_files, text data in flex_fields), a save that only touches text fields will wipe out the file data. It only works today because callers like validate_with_checker() happen to merge everything back into flex_fields before saving.

1. Storage split only, consumption still merged

Status: Partially addressed
Notes: Storage is split; some consumers still merge/read combined data by design.


2. Use JSONField instead of BinaryField for flex_files

Status: Open
Notes: Still BinaryField; encode/decode layer remains.

3. Checksum uses only first 8KB (FLEX_FILES_PREFIX)

Status: Resolved
Notes: Full flex_files content is now hashed (runtime + migration helper).


4. Implicit normalization in save() is fragile

Status: Open (design trade-off)
Notes: normalize_flex_storage() is still called from save().


5. _checker_file_fields() silently swallows errors

Status: Mostly resolved
Notes: Broad exception swallowing removed from helper; model-level catch narrowed to expected cases.


6. Individual admin view still merges all base64 and can crash browser

Status: Addressed
Notes: Change-form initials now use flex_fields only; no forced full merge for form init.


7. Inconsistent application: Kobo individual vs household split

Status: Resolved
Notes: Both individual and household Kobo paths now split into flex_fields + flex_files.


8. Many code paths still read flex_fields directly

Status: Open
Notes: Needs follow-up standardization (get_flex_value/combined access policy).


9. Field-type distinction should live in hope_flex_fields library

Status: Open (architectural decision)
Notes: Still implemented in country workspace layer.


10. Migration does row-by-row updates (no batching/logging)

Status: Open (minor/perf)
Notes: Functional, but not optimized for large volumes.


11. normalize_flex_storage can drop existing file data

Status: Resolved
Notes: Existing flex_files map is preserved and overlaid, not filtered out.

src/country_workspace/utils/flex_fields.py
```diff
@@ -14,7 +15,90 @@
...
+
+    Keys are sorted so the resulting bytes are deterministic (stable checksums).
+    """
+    if not value:
+        return None
+    return pickle.dumps(dict(sorted(value.items())), protocol=pickle.HIGHEST_PROTOCOL)
+
+
+def get_checker_file_fields(checker: DataChecker | None) -> set[str]:
+    """Return the set of file-field names for a checker.
```
datamik: We just need to update flex_fields library version and use methods from there. I think this check is extra

src/country_workspace/utils/flex_fields.py
```diff
@@ -14,7 +15,90 @@
...
+
+    The blob only ever holds file-typed keys, so no field-name filtering (and
+    therefore no checker lookup) is needed here.
+    """
+    merged = dict(flex_fields or {})
+    merged.update({field_name: value for field_name, value in decode_flex_files_blob(flex_files).items() if value})
+    return merged
+
+
+def split_flex_payload(payload: dict[str, Any], file_fields: set[str]) -> tuple[dict[str, Any], dict[str, Any]]:
```
datamik: Don't we have implemented this method as well in flex_fields library?

src/country_workspace/models/base.py
```diff
@@ -106,6 +114,56 @@ def save(
             update_fields=update_fields,
         )
 
+    def _checker_file_fields(self) -> set[str]:
+        try:
+            checker = self.checker
```
saxix: why this? What is the case when 'self.checker == null'

src/country_workspace/utils/flex_fields.py
```diff
@@ -14,7 +15,90 @@
...
+
+
+def encode_flex_files_blob(value: dict[str, Any]) -> bytes | None:
+    """Pickle a ``{field_name: content}`` mapping for storage in ``flex_files``.
+
+    Keys are sorted so the resulting bytes are deterministic (stable checksums).
+    """
+    if not value:
+        return None
+    return pickle.dumps(dict(sorted(value.items())), protocol=pickle.HIGHEST_PROTOCOL)
```
saxix: this should be in the library. Let use msgpack instead of standard pickle

src/country_workspace/models/base.py
saxix, line 85: flex_fields should be a dict of `{field_name: field_value}` it is better to use `default={}, blank=True, null=False`

saxix (changes requested):

datamik:
1. Files are still encoded to bse64 before pickling. The idea of using BinaryField was to store files in bytes
2. This PR does not touch files imported via RDI or Aurora
3. normalize_flex_storage still happens in save() method. Also it "fixes" the files in the flex_fields situation which should not happen at all.
4. New version of hope_flex_fields is not used

src/country_workspace/models/base.py
arsen-vs: Current code is JSONField(default=dict, blank=True) which is functionally equivalent for null behavior and is the safer Django pattern (avoids mutable default literal).

src/country_workspace/utils/import_flow/transformations.py
```diff
@@ -24,10 +24,10 @@ def _apply_transformer(qs: "QuerySet[Beneficiary]", transformer: Transformer | N
         transformed = transformer.apply(dict(current))
 
         if transformed != current:
-            record.flex_fields = transformed
+            updated = record.apply_flex_payload(transformed, preserve_existing_files=False)
```
vitali-yanushchyk-valor: Although `flex_files` is stored separately, this call explicitly rebuilds it. `transformed` is based on `record.flex_fields` only, and `preserve_existing_files=False` prevents `apply_flex_payload()` from loading the existing file blob. If the transformed payload contains no file fields, `flex_files` is therefore replaced with `None`.

src/country_workspace/models/base.py
vitali-yanushchyk-valor, line 119: `normalize_flex_storage()` is never called, so file values can remain in `flex_fields` instead of being moved to `flex_files`. This can leave conflicting values in both fields. What about call it before `update_checksum()` so the checksum is calculated from the normalized data.

src/country_workspace/utils/flex_fields.py
```diff
@@ -13,8 +16,104 @@
...
+            if (public_value := to_public_flex_file_value(value))
+        }
+    )
+    return merged
+
+
+def split_flex_payload(checker: DataChecker | None, payload: dict[str, Any]) -> tuple[dict[str, Any], dict[str, Any]]:
+    if checker is None:
+        return dict(payload), {}
+    split = checker.split_data(payload)
```
vitali-yanushchyk-valor: `checker.split_data()` resolves the file field names through ORM queries on every call. This helper is used inside the Kobo, XLSX, and Aurora record loops, so the same DataChecker configuration is loaded once per record. What about resolve the file field names once before each import loop and reuse that set when splitting payloads.

src/country_workspace/migrations/0055_split_file_flex_fields.py
```diff
@@ -0,0 +1,157 @@
...
+
+def _checker_file_field_names(apps: StateApps) -> dict[int, set[str]]:
+    FieldDefinition = apps.get_model("hope_flex_fields", "FieldDefinition")
+    FlexField = apps.get_model("hope_flex_fields", "FlexField")
+    DataCheckerFieldset = apps.get_model("hope_flex_fields", "DataCheckerFieldset")
+
+    file_definition_ids = {
+        definition.id
+        for definition in FieldDefinition.objects.only("id", "field_type").iterator()
+        if "Base64ImageField" in str(definition.field_type)
```
vitali-yanushchyk-valor: very minor: This works for the current `Base64ImageField`, but using the same `is_file` check as runtime would keep this logic consistent with `DataChecker`.

src/country_workspace/migrations/0055_split_file_flex_fields.py
```diff
@@ -0,0 +1,157 @@
...
+    ):
+        fields_by_fieldset[flex_field.fieldset_id].add(flex_field.name)
+
+    checker_fields: dict[int, set[str]] = defaultdict(set)
+    for member in DataCheckerFieldset.objects.only("checker_id", "fieldset_id", "prefix").iterator():
+        names = fields_by_fieldset.get(member.fieldset_id, set())
+        if not names:
+            continue
+        prefix = member.prefix or ""
+        checker_fields[member.checker_id].update(f"{prefix}{name}" for name in names)
```
vitali-yanushchyk-valor: This does not handle the supported `%s` prefix format. For example, `national_id_%s` + `photo` is `national_id_photo` at runtime, but the migration looks for `national_id_%sphoto`. Such file fields will stay in `flex_fields` instead of being moved to `flex_files`.

src/country_workspace/migrations/0063_split_file_flex_fields.py
```diff
@@ -0,0 +1,157 @@
...
+
+    fields_by_fieldset: dict[int, set[str]] = defaultdict(set)
+    for flex_field in (
+        FlexField.objects.filter(definition_id__in=file_definition_ids).only("fieldset_id", "name").iterator()
+    ):
+        fields_by_fieldset[flex_field.fieldset_id].add(flex_field.name)
+
+    checker_fields: dict[int, set[str]] = defaultdict(set)
+    for member in DataCheckerFieldset.objects.only("checker_id", "fieldset_id", "prefix").iterator():
+        names = fields_by_fieldset.get(member.fieldset_id, set())
```
vitali-yanushchyk-valor, line 115: very minor: This works with the current `DataCheckers`, but it does not include fields inherited through `Fieldset.extends`, while `DataChecker.get_fields() `does. It may be worth keeping the migration logic consistent with the runtime behavior.

src/country_workspace/workspaces/admin/batch/reprocessing.py
```diff
@@ -101,10 +101,10 @@ def _apply_import_processor(
         flex_fields |= {
             field: value for field, attr in preserved.items() if (value := getattr(record, attr, None)) is not None
         }
-    record.flex_fields = flex_fields
+    updated = record.apply_flex_payload(flex_fields, preserve_existing_files=True)
```
vitali-yanushchyk-valor: Could you check this case during reprocessing? If a wrong mapping like `passport_photo=photo` is later corrected to `passport_photo=national_passport_image`, it looks like the old `photo` may remain in `flex_files`. Would it make sense to use `preserve_existing_files=False` here since reprocessing rebuilds the record from `raw_data`?

src/country_workspace/workspaces/admin/cleaners/bulk_update.py
```diff
@@ -217,7 +217,7 @@ def create_bulk_update_template(queryset: "QuerySet[Beneficiary]", program: Prog
         fmt = lambda v: ", ".join(map(str, v)) if isinstance(v, list | tuple) else str(v if v is not None else "")
         for row, record in enumerate(queryset.iterator(chunk_size=ITERATOR_CHUNK_SIZE), 1):
             for col, fld in enumerate(columns):
-                value = getattr(record, fld, record.flex_fields.get(fld))
+                value = getattr(record, fld, record.get_flex_value(fld))
```
vitali-yanushchyk-valor: Previously the fallback only read `flex_fields`, but `get_flex_value()` may now access deferred `flex_files` even for `id/version`, causing an extra query per record. Could we only call it when the field isn't available directly on the model instance?

src/country_workspace/migrations/0055_split_file_flex_fields.py
arsen-vs: Solved

src/country_workspace/migrations/0055_split_file_flex_fields.py
arsen-vs: Solved

src/country_workspace/migrations/0063_split_file_flex_fields.py
arsen-vs: Solved

src/country_workspace/workspaces/admin/batch/reprocessing.py
arsen-vs: Solved

src/country_workspace/workspaces/admin/cleaners/bulk_update.py
arsen-vs: Solved

src/country_workspace/migrations/0063_split_file_flex_fields.py
```diff
@@ -0,0 +1,205 @@
...
+    def _resolve_effective_fields(fieldset_id: int, trail: set[int]) -> dict[str, bool]:
+        if fieldset_id in resolved:
+            return resolved[fieldset_id]
+        if fieldset_id in trail:
+            return {}
+
+        effective: dict[str, bool] = dict(fields_by_fieldset.get(fieldset_id, []))
+        parent_id = extends_by_fieldset.get(fieldset_id)
+        if parent_id:
+            effective.update(_resolve_effective_fields(parent_id, trail | {fieldset_id}))
```
vitali-yanushchyk-valor, line 93: minor: This now handles inherited fields, but the override order is reversed. `Fieldset.get_fields()` lets local fields override parent fields, while here `effective.update(parent) `makes the parent win. This can misclassify overridden fields as file or non-file.

src/country_workspace/workspaces/admin/cleaners/bulk_update.py
vitali-yanushchyk-valor: This still seems unresolved: `get_flex_value()` is evaluated eagerly here, so accessing deferred `flex_files` causes an extra query per record.

src/country_workspace/utils/flex_fields.py
```diff
@@ -13,8 +16,109 @@
...
+    text_fields = dict(split.get("fields", {}))
+    raw_file_values = dict(split.get("files", {}))
+
+    file_values = {
+        key: to_storage_flex_file_value(value)
+        for key, value in raw_file_values.items()
+        if value not in (None, "") and not (isinstance(value, str) and not value.strip())
```
vitali-yanushchyk-valor: Is this intentional? When a file field is explicitly set to an empty value, e.g. `photo=""`, the existing value in `flex_files` is preserved. Should an explicit empty value remove the previously stored file instead?

src/country_workspace/models/base.py
```diff
@@ -106,6 +114,61 @@ def save(
...
+        row is written, while existing file values are preserved.
+        """
+        if update_fields is not None and not (set(update_fields) & {"flex_fields", "flex_files"}):
+            return update_fields
+
+        if self.checker is None:
+            return update_fields
+
+        text_fields, new_file_values = split_flex_payload(self.checker, self.flex_fields or {})
+        existing_file_values = self.get_flex_files_map()
```
vitali-yanushchyk-valor: minor: Could this become an issue if a field changes from a file field to a text field? Its old value would remain in `flex_files`, and `get_combined_flex_fields()` could then return that old file value instead of the current text value. Should we keep only values for fields that are still file fields?

src/country_workspace/contrib/aurora/import_processing.py
```diff
@@ -204,28 +219,45 @@ def create_individual(
...
         mapping_id=config.get("individual_mapping_id"),
     )
+    transformed = individual_row_processor(row)
+    text_fields, file_values = split_flex_payload(
+        batch.program.individual_checker,
+        transformed,
+    )
     extras.setdefault("household", None)
     return Individual.objects.create(
         batch_id=batch.pk,
```
vitali-yanushchyk-valor: Could we pass` batch=batch` here instead of `batch_id`? The new save-time normalization accesses` self.checker` through `self.batch.program`, so using only `batch_id` may cause `Batch`, `Program`, and `DataChecker` to be loaded again for every created record. The same applies to` create_household() `below.

src/country_workspace/datasources/rdi/processors.py
```diff
@@ -128,13 +130,16 @@ def process_households(sheet: Sheet, job: AsyncJob, batch: Batch, config: Config
             raise SheetProcessingError(SheetName.HOUSEHOLDS, household_key)
         originating_id = get_xlsx_originating_id(file_name, household_key, epoch=epoch_ms)
         try:
+            transformed = transform_row(row)
+            text_fields, file_values = split_flex_payload(checker, transformed)
             mapping[household_key] = cast(
                 "Household",
                 Household.objects.create(
                     batch_id=batch.pk,
```
vitali-yanushchyk-valor: Could we pass `batch=batch` here as well? We already have the batch and checker loaded before the loop, but save-time normalization may load them again for each record when only `batch_id` is set. The same applies to the individual creation below.

src/country_workspace/contrib/kobo/sync.py
```diff
@@ -188,13 +186,19 @@ def create_individuals(  # noqa: PLR0913
...
+    )
     for idx, raw_individual in enumerate(submission.get(config["individual_records_field"], []), start=1):
         if job:
             job.ensure_not_cancelled(refresh=True)
-        individual_fields = build_individual_processor(
-            batch.program,
-            individual_mapping_id,
-        )(raw_individual)
+        individual_fields = processor(raw_individual)
+        text_fields, file_values = split_flex_payload(
```
vitali-yanushchyk-valor: minor: Could we move this split into the `else` branch that creates a regular Individual? External collectors still pass the original individual_fields to `get_or_create_collector()`, so text_fields and file_values are unused for that branch.

src/country_workspace/migrations/0063_split_file_flex_fields.py
arsen-vs: Now parent is resolved first, then local fields override:

src/country_workspace/utils/flex_fields.py
arsen-vs: Explicit empty file should clear existing file

split_flex_payload() still excludes empty file values (as before), but clearing is now handled in model write paths:
In apply_flex_payload() (models/base.py), it reads raw checker file data and removes keys where value is empty (None or blank string) from existing flex_files.
Same logic is in normalize_flex_storage().
Net effect: photo="" removes previously stored file value.

src/country_workspace/models/base.py
arsen-vs: Stale file value after field type change

In both apply_flex_payload() and normalize_flex_storage(), existing flex_files are filtered to only keys currently in checker.get_file_field_names().
So old file keys for fields that are no longer file fields are dropped and won’t override text values.

vitali-yanushchyk-valor:
https://github.com/unicef/hope-country-workspace/blob/4693a598c79beefccc233fe3d0752702fd01b69a/src/country_workspace/workspaces/admin/hh_ind.py#L298
Could this cause existing file values to be lost when editing a record? File fields are now stored in `flex_files`, so they are not included in these `initials`. If no new file is uploaded, the form may return an empty value for the file field, which could then clear the existing file during save normalization. Would it make sense to build the initials from `obj.get_combined_flex_fields()` here?

vitali-yanushchyk-valor:
https://github.com/unicef/hope-country-workspace/blob/4693a598c79beefccc233fe3d0752702fd01b69a/src/country_workspace/workspaces/admin/cleaners/mass_update.py#L122
Could this cause existing file fields to be treated as missing? After this change, file values are stored in `flex_files`, so a field such as `photo` would no longer satisfy this condition. If **“Create missing fields”** is not selected, operations like `set` or `set null` could then be silently skipped for an existing file. Could we handle both flex storage locations here?

vitali-yanushchyk-valor:
https://github.com/unicef/hope-country-workspace/blob/4693a598c79beefccc233fe3d0752702fd01b69a/src/country_workspace/models/base.py#L163
Could this cause repeated queries in operations that save many records one by one, such as Mass Update, picture import, etc? `normalize_flex_storage() `runs for every record saved with flex `fields`, and resolving `self.checker` here may load the same program/checker data again for each record. Could we reuse the checker across the operation?

vitali-yanushchyk-valor:
https://github.com/unicef/hope-country-workspace/blob/4693a598c79beefccc233fe3d0752702fd01b69a/src/country_workspace/workspaces/admin/cleaners/calculate_checksum.py#L10
Could this now cause an N+1? `queryset.all()` uses the custom `ValidableQuerySet.all()`, which defers `flex_files`, but `record.save()` immediately needs that field for normalization and checksum calculation. This may trigger one extra `flex_files` query per record. Would iterating over `queryset` directly instead of `queryset.all()` here be sufficient to avoid that?

vitali-yanushchyk-valor:
https://github.com/unicef/hope-country-workspace/blob/4693a598c79beefccc233fe3d0752702fd01b69a/src/country_workspace/workspaces/admin/cleaners/regex.py#L41
Could file fields be excluded from this action? After the split, an existing file value is stored in `flex_files`, while this action still reads and updates only `flex_fields`. Selecting an existing file field would therefore treat it as missing and can add a shadow value to `flex_fields`, while `bulk_update() `bypasses the new storage normalization. Would it make sense to limit this action to non-file fields? The `Concatenate` action seems to have the same issue.

src/country_workspace/workspaces/admin/hh_ind.py
vitali-yanushchyk-valor: minor: Could this also consider `flex_files`? File changes are now tracked by `pghistory`, but this comparison still reads only `entry.flex_fields`. This may cause file replacements or clears to be missing from the displayed history, and older events that still have file values in `flex_fields` could also appear as if the file was cleared. Would using the combined flex payload for each event make sense here?

src/country_workspace/utils/flex_fields.py
```diff
@@ -13,24 +18,181 @@
...
+
+def split_flex_storage(
+    checker: DataChecker | None,
+    payload: dict[str, Any],
+    file_field_names: set[str] | None = None,
+) -> FlexStorageSplit:
+    if checker is None:
+        return FlexStorageSplit(dict(payload), {}, set())
+
+    split = checker.split_data(payload, file_field_names=file_field_names)
```
datamik, line 115: Can we pick better naming? "split" does not give any meaningful information

src/country_workspace/models/base.py
```diff
@@ -35,6 +44,13 @@ class ValidableQuerySet(BaseQuerySet):
     def all(self) -> "QuerySet[Model, Model]":
         return super().all().defer("flex_files")
 
+    def with_flex_storage(self) -> "QuerySet[Model, Model]":
```
datamik, line 47: How the name of this method is related to what it actually does?

src/country_workspace/migrations/0063_split_file_flex_fields.py
```diff
@@ -0,0 +1,206 @@
...
+
+def _decode_blob(blob: bytes | memoryview | bytearray | None) -> dict[str, Any]:
+    if not blob:
+        return {}
+    raw = bytes(blob) if isinstance(blob, memoryview | bytearray) else blob
+    try:
+        parsed = msgpack.unpackb(raw, raw=False, strict_map_key=False)
+    except (msgpack.ExtraData, msgpack.FormatError, msgpack.StackError, ValueError, TypeError):
+        try:
+            parsed = json.loads(raw.decode("utf-8"))
```
datamik, line 28: Do we need that fallback at all?

src/country_workspace/migrations/0063_split_file_flex_fields.py
```diff
@@ -0,0 +1,206 @@
...
+        files_map = _decode_blob(record.flex_files)
+        changed = False
+
+        for field_name in file_fields:
+            if field_name not in flex_fields:
+                continue
+
+            value = flex_fields.pop(field_name)
+            if isinstance(value, str) and value.strip() and field_name not in files_map:
+                files_map[field_name] = value
```
datamik, line 167: This will store file in base64 right? i think we first need to decode it

datamik:
This will be implemented using a different approach.

PR #419, opened 2026-06-30
Update RDP lifecycle after cloning removal

vitali-yanushchyk-valor:
## Summary

Updates the RDP lifecycle after removing cloning.

## Changes

* Removed RDP cloning.
* Recreate HOPE RDI on push retry: if an RDP already has `hope_rdi_id`, it means a previous push attempt created an RDI but failed while sending beneficiaries; in this case, CW asks HOPE Core to delete that existing unfinished RDI before creating a fresh one.
* Added compact RDP operation audit log.
* Prevent dedup settings changes after a successful RDP, or while an open RDP is using/running deduplication.
* Prevent cancellation while push or deduplication is queued/running, and prevent starting another push while the same RDP push is already queued or running.


src/country_workspace/contrib/hope/push/orchestration.py
```diff
@@ -108,146 +71,111 @@ def create_rdp_core(job: AsyncJob) -> dict[str, Any]:
...
     policy = get_rdp_policy(rdp)
-    check = policy.claim_deduplication_check()
+    check = policy.deduplicate_check()
     if not check.allowed:
         return check, None
 
-    update_fields = ["is_dedup_settings_locked"]
-    rdp.is_dedup_settings_locked = True
+    with transaction.atomic():
+        locked = lock_rdp_for_update(pk=rdp_id)
```
datamik, line 89: Maybe locked_rdp is better name

datamik (approved):

roma-valor (approved):

PR #421, opened 2026-07-01
feat: Add rsa decryption flow

roma-valor:
# Legal Boilerplate

Look, I get it.
under HOPE's choice of terms.

### DIFF
--- a/src/country_workspace/contrib/aurora/admin/registration.py
+++ b/src/country_workspace/contrib/aurora/admin/registration.py
@@ -36,4 +36,4 @@ class RegistrationAdmin(BaseModelAdmin):
         return False
 
     def has_change_permission(self, request: HttpRequest, obj: Registration | None = None) -> bool:
-        return request.user.is_superuser
+        return request.user.has_perm("aurora.change_registration")
--- a/src/country_workspace/contrib/aurora/admin/registration.py
+++ b/src/country_workspace/contrib/aurora/admin/registration.py
@@ -3,12 +3,14 @@ from django.contrib import admin
 from django.http import HttpRequest
 
 from country_workspace.admin.base import BaseModelAdmin
+from country_workspace.contrib.aurora.forms import RegistrationAdminForm
 from country_workspace.contrib.aurora.models import Registration
 
 
 @admin.register(Registration)
 class RegistrationAdmin(BaseModelAdmin):
-    list_display = ("name", "project", "active", "last_synced")
+    form = RegistrationAdminForm
+    list_display = ("name", "project", "active", "has_decryption_key", "last_synced")
     list_filter = (
         ("project", AutoCompleteFilter),
         "active",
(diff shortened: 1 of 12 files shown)


src/country_workspace/contrib/aurora/admin/registration.py
```diff
@@ -17,12 +19,21 @@ class RegistrationAdmin(BaseModelAdmin):
...
+        if obj is None:
+            return super().get_readonly_fields(request, obj)
+        return tuple(field.name for field in Registration._meta.fields if field.name not in {"id", "rsa_private_key"})
+
     def has_add_permission(self, request: HttpRequest) -> bool:
         return False
 
     def has_change_permission(self, request: HttpRequest, obj: Registration | None = None) -> bool:
-        return False
+        return request.user.is_superuser
```
datamik: Perhaps we need to check for the registration specific permissions. So that some users without superuser privileges is be able to add RSA key.

src/country_workspace/contrib/aurora/import_processing.py
```diff
@@ -33,12 +35,43 @@ class ImportResult(NamedTuple):
     households: int = 0
 
 
+def _load_rsa_private_key(registration_reference_pk: str) -> str:
+    registration = Registration.objects.filter(reference_pk=int(registration_reference_pk)).first()
```
vitali-yanushchyk-valor: Could we scope this lookup by the current program as well?

This is not critical, but the import form already limits registrations to `project__program=program`, so I think the private key lookup should use the same scope for consistency.

src/country_workspace/contrib/aurora/import_processing.py
```diff
@@ -33,12 +35,43 @@ class ImportResult(NamedTuple):
...
+    if registration is None:
+        return ""
+    return (registration.rsa_private_key or "").strip()
+
+
+def prepare_record(record: Mapping[str, Any], private_key: str) -> dict[str, Any]:
+    result = dict(record)
+    fields = result.get("fields")
+    if fields is None or isinstance(fields, dict):
```
vitali-yanushchyk-valor: Tiny cleanup: should we check `fields` with `Mapping` instead of `dict` here?

roma-valor: resolved

src/country_workspace/contrib/aurora/import_processing.py
roma-valor: resolved

src/country_workspace/contrib/aurora/admin/registration.py
roma-valor: resolved

src/country_workspace/contrib/aurora/import_processing.py
```diff
@@ -33,12 +35,57 @@ class ImportResult(NamedTuple):
...
+    if "payload" in result:
+        payload = result.pop("payload")
+        if not private_key:
+            msg = (
+                f"Record {result.get('pk')}: encrypted Aurora registration requires an RSA private key "
+                "on the CW Registration"
+            )
+            raise ImportError(msg)
+        if not isinstance(payload, Mapping) or "encryption" not in payload:
+            msg = f"Record {result.get('pk')}: Aurora refused to return the encrypted payload ({payload})"
```
vitali-yanushchyk-valor, line 60: minor: I’d avoid interpolating the full payload into this error message. It can be large and noisy in job errors/logs.

vitali-yanushchyk-valor (approved):

datamik (approved):

PR #422, opened 2026-07-08
Feature/320208 use shared blob storage for file transfer

datamik:
[AB#320208](https://unicef.visualstudio.com/4e044e8d-bc28-4768-8074-f80ff6e4a0e5/_workitems/edit/320208)

Store images in shared HOPE blob storage instead of embedding base64 payloads when pushing data to external HOPE services.

Add HOPE_STORAGE (Azure Blob, Azurite locally) with deploy checks and an init_hope_storage management command to create the container on startup.
Add blob_hashes field on Household/Individual to track uploaded image content hashes, enabling upload/reuse/delete reconciliation instead of re-uploading unchanged images.
Reconcile flex-field images against blob storage during push/dedup, replacing base64 values with blob keys before sending to HOPE.


src/country_workspace/services/hope_blob.py
```diff
@@ -0,0 +1,104 @@
...
+        if only is not None and name not in only:
+            continue
+        value = record.flex_fields.get(name)
+        if not is_data_uri(value):
+            continue
+        present.add(name)
+        key = record.hope_blob_key(name)
+        new_hash = hashlib.sha256(value.encode()).hexdigest()
+        if hashes.get(name) != new_hash:
+            if HOPE_STORAGE.exists(key):
```
saxix: can we simply ovverride existing if exists? What is the case where the image exists ?

src/country_workspace/services/hope_blob.py
```diff
@@ -0,0 +1,104 @@
...
+
+def substitute_row_images(record: "Validable", row: dict, paths: dict[str, str]) -> dict:
+    def repl(name: str, value: Any) -> Any:
+        return paths.get(name, value) if is_data_uri(value) else value
+
+    out = {}
+    for key, value in row.items():
+        if isinstance(value, list):
+            out[key] = [
+                {k: repl(f"{(it.get('type') or key)}_{k}", v) for k, v in it.items()} if isinstance(it, dict) else it
```
saxix: be safe. Let create a structure or TypedDict for `it` to be sure `type` exists. Let mypy do the job for us

src/country_workspace/storages.py
```diff
@@ -1,8 +1,12 @@
...
+from django.utils.functional import SimpleLazyObject
 from django.utils.module_loading import import_string
 
 
 backend = import_string(settings.STORAGES["media"]["BACKEND"])
 options = settings.STORAGES["media"]["OPTIONS"]
 
 MEDIA_STORAGE = backend(**options)
+
+HOPE_STORAGE = SimpleLazyObject(lambda: storages["hope"])
```
saxix: add a note that SimpleLazyObject is safe because we have check.py.

src/country_workspace/management/commands/init_hope_storage.py
```diff
@@ -0,0 +1,22 @@
...
+        storage = storages["hope"]
+        client = getattr(storage, "client", None)
+        if client is None or not hasattr(client, "create_container"):
+            self.stdout.write("STORAGES['hope'] is not an Azure backend; nothing to do.")
+            return
+        try:
+            client.create_container()
+            self.stdout.write(self.style.SUCCESS("Created HOPE blob container."))
+        except Exception as e:  # noqa: BLE001
+            self.stdout.write(f"HOPE blob container not created ({e.__class__.__name__}); assuming it already exists.")
```
saxix: too broader catch. make difference failed creation, existing storage and unkonwn error

saxix (changes requested):

src/country_workspace/services/hope_blob.py
```diff
@@ -0,0 +1,104 @@
...
+    image_fields: Sequence[str],
+    only: set[str] | None = None,
+) -> dict[str, str]:
+    """Reconcile record images with the shared HOPE blob storage.
+
+    Raises BlobStorageError when the storage backend is unreachable or fails.
+    """
+    try:
+        return _sync_record_blobs(record, image_fields, only)
+    except (AzureError, OSError) as e:
```
vitali-yanushchyk-valor: minor: This wraps Azure/storage failures, but malformed Base64 from `decode_data_uri()` can still escape this path and crash the job. Should we decode with `validate=True` and wrap `binascii.Error` as a blob sync failure too?

vitali-yanushchyk-valor (approved):

vitali-yanushchyk-valor (approved):

docs/src/config/storage.md
```diff
@@ -0,0 +1,65 @@
...
+storage.save(key, content_file)
+storage.open(key, "rb")
+storage.exists(key)
+storage.delete(key)
+```
+
+Handle missing blobs with `azure.core.exceptions.ResourceNotFoundError` (and `FileNotFoundError` for local filesystem storage in tests).
+
+!!! note "Blob keys, not file bytes"
+    The database stores blob **keys/paths as plain strings**, not the file contents and not a `FileField`. Agree a key-prefix convention with HOPE core (for example `households/{pk}/photo.jpg`) so writes from different services do not collide.
```
saxix: is this url correct. Does not seem to match the code

saxix (approved):

PR #439, opened 2026-07-28
feat: Deduplicate kobo external collectors program-wide at import

roma-valor:
## Problem

When importing from Kobo, an external collector (`relationship == NON_BENEFICIARY`)
who appears in multiple submissions was created as a **separate `Individual` for every
submission**, each with the `household` FK set — i.e. as a regular household member.

Pushed to HOPE via the lax endpoints, this produced data that violates HOPE's own
invariant (`household IS NULL ⟺ NON_BENEFICIARY`): HOPE's legacy direct kobo import
creates such people household-less and links them to households only through
`IndividualRoleInHousehold` rows. As a member with `NON_BENEFICIARY` relationship,
the collector was excluded from household size/demographic recalculations yet still
listed among household members — inconsistent state in HOPE.

## Solution

Deduplicate external collectors **program-wide at import time** in HCW, so HOPE
receives exactly what its own legacy import would have produced: one household-less
`Individual` + one collector role row per household.

### Import (`contrib/kobo/sync.py`)

- `create_individuals()` routes `NON_BENEFICIARY` records through the new
  `get_or_create_collector()` (`utils/import_flow/collector_identity.py`):
  - matches on the new indexed `Individual.identity_hash` — sha256 over the same
    identity fields HOPE's `get_hash_key` uses (names, sex, birth_date, phones),
    scoped per program via `batch__program`;
  - first occurrence wins: the existing record is reused as-is, later occurrences'
    data is **not** merged (documented limitation, possible follow-up);
  - the collector is created with `household=None`; linkage to households keeps
    working through the existing `primary_collector_id` / `alternate_collector_id`
    flex-field refs (`set_roles_and_relationships` unchanged);
  - a transaction-scoped Postgres advisory lock per `(program, hash)` serializes
    concurrent import jobs (imports run in parallel, so per-job dedup was not an
    option);
  - collectors without any identity data (hash `None`) are **not** deduplicated —
    current behavior preserved.

### Migrations

- `0059` — adds `Individual.identity_hash` (indexed, nullable).
- `0060` — remove and re-create Individual trigger
  imports can match collectors imported before this change.

### Push (`contrib/hope/push/`)

Household-less collectors are not selected by `household_id__in`, so without this
change they would never be pushed and `map_role_value` would fail to resolve the
role refs:

- new `qs_individuals_for_push()` = household members ∪ collectors referenced by
  the pushed households' role ref fields — used by the push-individuals step;
- `preflight_errors()` checks those collectors too (validity, already-in-another-RDP);
- `_mark_rdp_beneficiaries_removed` intentionally unchanged: a shared collector may
  be referenced by other RDPs, so RDP removal does not toggle its `removed` flag.

## Out of scope

- Legacy data cleanup (existing duplicate collector copies stay as-is; the backfill
  only adds hashes, no record merging).
- XLSX/RDI datasource path (`datasources/rdi/processors.py`) — kobo only for now.
- Merging documents/flex fields of duplicate occurrences.

## Testing

- New: `tests/utils/import_flow/test_collector_identity.py` — hash normalization,
  program-wide reuse, per-program scoping, removed-record handling, no-identity
  fallback.
- New: kobo sync tests — collector routing, cross-household/cross-batch dedup with
  role refs pointing at the single record, member+collector mix in one submission.
- New: push repository tests — role-ref pk extraction, push queryset union,
  preflight coverage of referenced collectors.
- `ruff check` / `ruff format --check` clean.


vitali-yanushchyk-valor:
Could you please update the project documentation in this PR to reflect this change?

src/country_workspace/contrib/kobo/sync.py
```diff
@@ -175,17 +180,32 @@ def create_individuals(  # noqa: PLR0913
...
-                flex_fields=individual_fields,
+                individual_fields=individual_fields,
                 raw_data=raw_individual,
-            ),
-        )
-    household.program.individuals.bulk_create(individuals)
+                originating_id=ind_originating_id,
+                name=name,
+            )
+            individuals.append(individual)
```
vitali-yanushchyk-valor: Is it valid from a business perspective for the same collector to be PRIMARY for one household and ALTERNATE for another? If so, when the collector is reused, its role still comes from the first submission that created the shared collector record, not from the current submission. As a result, `set_roles_and_relationships()` may assign the wrong collector reference to a later household.

src/country_workspace/contrib/kobo/sync.py
vitali-yanushchyk-valor, line 505: How are collectors with `household=None` validated after Kobo import? `create_validation_jobs()` schedules only the households from this queryset, and `Household.validate_with_checker()` validates only `self.members`. Referenced collectors therefore can keep `last_checked=None`, and the RDP preflight check rejects them as unvalidated.

src/country_workspace/utils/import_flow/collector_identity.py
```diff
@@ -0,0 +1,93 @@
...
+                batch__program=program, identity_hash=identity_hash, household=None, removed=False
+            )
+            .order_by("id")
+            .first()
+        )
+        if existing is not None:
+            return existing, False
+    return (
+        Individual.objects.create(
+            batch=batch,
```
vitali-yanushchyk-valor, line 84: The shared collector keeps the `batch` from the import where it was first created, even when households from later batches reuse it. Cleaning up that first Batch deletes the collector and leaves its PK in the `flex_fields` of households from later batches, pointing to a record that no longer exists. Reprocessing has the same ownership problem: reprocessing the first Batch may change the shared collector, while reprocessing later batches does not include it.

src/country_workspace/utils/import_flow/collector_identity.py
```diff
@@ -0,0 +1,93 @@
...
+    every later import; its data is never modified. The record is created with
+    ``household=None`` — linkage to households happens exclusively through the
+    primary/alternate collector role reference fields.
+    """
+    identity_hash = compute_collector_hash(individual_fields)
+    if identity_hash is not None:
+        _lock_collector(program.pk, identity_hash)
+        existing = (
+            Individual.objects.filter(
+                batch__program=program, identity_hash=identity_hash, household=None, removed=False
```
vitali-yanushchyk-valor, line 75: Could this leave some collector links unresolved? This query may return a collector created in an earlier Batch, while `sync_collector_links()` looks only at Individuals from the current Batch. In that case, the original `collector_id` in `Individual.flex_fields` may remain unchanged instead of being converted to the collector PK.

src/country_workspace/migrations/0060_backfill_individual_identity_hash.py
```diff
@@ -0,0 +1,57 @@
...
+    values = [_normalize(flex_fields.get(field)) for field in COLLECTOR_HASH_FIELDS]
+    if not any(values):
+        return None
+    return hashlib.sha256(";".join(values).encode()).hexdigest()
+
+
+def backfill_identity_hash(apps: StateApps, schema_editor: BaseDatabaseSchemaEditor) -> None:
+    Individual = apps.get_model("country_workspace", "Individual")
+    for individual in (
+        Individual.objects.filter(flex_fields__relationship="NON_BENEFICIARY", identity_hash__isnull=True)
```
vitali-yanushchyk-valor: Could this backfill have any effect on legacy collectors? Before this change, `NON_BENEFICIARY` Individuals were stored with a household, but the new lookup only reuses collectors with `household=None`. This migration sets `identity_hash` but leaves the household unchanged, so a later import of the same collector may create a new record instead of reusing the legacy one.

src/country_workspace/utils/import_flow/collector_identity.py
```diff
@@ -0,0 +1,93 @@
...
+    individual_fields: dict[str, Any],
+    raw_data: dict[str, Any],
+    originating_id: str,
+    name: str = "",
+) -> tuple[Individual, bool]:
+    """Return the single program-wide Individual for an external collector.
+
+    External collectors (relationship == NON_BENEFICIARY) are deduplicated
+    program-wide by identity hash: the first occurrence wins and is reused by
+    every later import; its data is never modified. The record is created with
```
vitali-yanushchyk-valor, line 66: Could transformers or batch reprocessing break this assumption? Both can update `Individual.flex_fields` without recalculating `identity_hash`. If one of `COLLECTOR_HASH_FIELDS` changes, a later import may calculate a different hash and create another collector instead of reusing this one.

src/country_workspace/migrations/0060_backfill_individual_identity_hash.py
roma-valor: removed

src/country_workspace/migrations/0060_remove_individual_update_update_and_more.py
```diff
@@ -0,0 +1,43 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("country_workspace", "0059_individual_identity_hash"),
+    ]
+
+    operations = [
+        pgtrigger.migrations.RemoveTrigger(
```
vitali-yanushchyk-valor, line 14: Was the legacy collector backfill intentionally removed? The PR description still says migration 0060 backfills `identity_hash`, but the current migration only updates the history model and trigger. Legacy collectors also still have a household, while the new lookup only reuses `household=None` records, so a later import may create a new collector instead of reusing the legacy one.

roma-valor: fixed

src/country_workspace/contrib/kobo/sync.py
vitali-yanushchyk-valor, line 234: How can external collector links be rebuilt during Kobo reprocessing?

Before this change, every successfully imported repeat row was stored in its own Household-linked `Individual.raw_data`, so reprocessing could process those records again and derive the Household role references. The repeat group was already excluded from `Household.raw_data`.

With program-wide collector reuse, later occurrences of the same external collector no longer create their own Individual, so those Household-specific raw rows are not stored anywhere. Reprocessing therefore has no source data to determine whether the collector was PRIMARY or ALTERNATE for the later Households that reuse it.

Could we keep the original repeat group in `Household.raw_data` while still excluding it from household processing, and use those raw rows during reprocessing through the same Kobo individual processing and collector deduplication flow as the initial import? Reprocessing would then need to clear and rebuild the collector links and correctly handle rows that change between a household member and `NON_BENEFICIARY`, without modifying a shared collector used by other Households.

src/country_workspace/utils/import_flow/transformations.py
```diff
@@ -27,7 +29,12 @@ def _apply_transformer(qs: "QuerySet[Beneficiary]", transformer: Transformer | N
             record.flex_fields = transformed
             record.last_checked = None
             record.errors = {}
-            record.save(update_fields=("flex_fields", "last_checked", "errors"))
+            update_fields = ["flex_fields", "last_checked", "errors"]
+            if isinstance(record, Individual) and transformed.get("relationship") == RELATIONSHIP_NON_BENEFICIARY:
+                # Keep the dedup key in sync with the stored identity fields,
+                record.identity_hash = compute_collector_hash(transformed)
```
vitali-yanushchyk-valor: Could this create a duplicate collector in a later Kobo import?

`get_or_create_collector()` looks for an existing collector using a hash calculated before the individual transformer runs. Later, post-processing recalculates and stores the hash from the transformed data. If the transformer changes an identity field in a way that changes the hash, the next import calculates the original pre-transform hash again and cannot find the existing collector, so it creates a new one.

Could we make the lookup and stored hash use the same data? For example, either apply the transformer before calculating the lookup hash and avoid transforming the created collector again, or treat the pre-transform hash as the stable lookup key and do not overwrite it during post-processing.

src/country_workspace/utils/import_flow/transformations.py
vitali-yanushchyk-valor, line 24: Could the transformer leave incorrect Household links?

The member/collector type, Household role references, and `collector_id` links may already be set when this transformer runs. If it changes `relationship`, `role`, or `collector_id`, those links are not updated. For example, an Individual changed to `NON_BENEFICIARY` would still belong to the Household, while a collector changed to a beneficiary would remain household-less.

Could these structural fields be finalized before the links are created, or could the links be rebuilt after transformation?

saxix (changes requested):
from DS / DS4-Flash:

1. Removed-collector inconsistency: push path (collector_pks_by_household_pks → qs_individuals_for_push) has no removed=False filter, while _external_collectors (household.py) and _get_individuals_queryset (beneficiary_reference) do. A removed collector would still be pushed (and its map_role_value refs resolved) but skipped in validation. Align the push path with the removed=False filter — or document why removal doesn't apply to shared collectors there (the PR says _mark_rdp_beneficiaries_removed is intentionally unchanged, but the preflight/push path itself is new).

2. Potential lock-order deadlock: one submission creates N collectors → N advisory locks acquired in record order within the same transaction. Two parallel imports processing the same collectors in opposite order → deadlock; Postgres aborts one submission, which fails the whole import run (watermark resumes later, so no data loss, but a hard job failure). Low risk, but consider ORDER BY on the lock acquisition order or documenting the retry semantics. Worth a one-line note.

3. Stale hash on relationship change: transformations.py/reprocessing.py recompute identity_hash only when relationship == NON_BENEFICIARY; a collector transformed to another relationship keeps its old hash (dedup lookup excludes it via household=None filter, so it's inert — but a later identical import would then create a duplicate). Consider clearing the hash when relationship leaves NON_BENEFICIARY.

4. individual_counter += len(individuals) counts reused collectors as imported — cosmetic only, but the ImportResult is user-visible.

5. Shared collector validated once per referencing household (chain(self.members.all(), external_collectors)): N+1-ish and writes errors on the shared record repeatedly. Acceptable, but a distinct()/once-per-program pass would be cleaner.

src/country_workspace/contrib/kobo/sync.py
roma-valor: What do you think about a lighter alternative: adding the role ref fields (primary_collector_id / alternate_collector_id) to PRESERVED_FLEX_FIELDS for KOBO households?
In practice this only freezes links pointing to household-less collectors. For regular members, _sync_kobo_household_refs() still re-derives and overwrites those fields during reprocessing, so member links keep updating as before. The preserved value only survives where the rebuild can't derive anything, which is exactly the external collector case.

src/country_workspace/utils/import_flow/transformations.py
roma-valor: we don't expect a transformer to change someone to NON_BENEFICIARY automatically. Transformers are for data cleanup, not for changing record types, and no current configuration does this.
To be safe, I'll add a check that blocks changes to relationship, role, and collector_id both in transformers and for manual edits in the admin, so these fields can't be changed after import.
If someone ever needs to change them, let's handle it in a separate PR with proper link rebuilding, this PR is already big enough. WDYT?

src/country_workspace/contrib/kobo/sync.py
vitali-yanushchyk-valor: That sounds like a good lighter solution. Preserving `primary_collector_id` and `alternate_collector_id` should be enough if changes to `relationship`, `role`, and `collector_id` are blocked in Mapping Importers during reprocessing, in transformers, and in manual admin edits.

src/country_workspace/utils/import_flow/transformations.py
vitali-yanushchyk-valor: That sounds good to me. Blocking changes to `relationship`, `role`, and `collector_id` in Mapping Importers during reprocessing, in transformers, and in manual admin edits should keep the links correct.

roma-valor:
@saxix regarding your comment:

1. The asymmetry is real, but I'd rather not add the filter:
    - The push path has never filtered `removed` - `qs_individuals_by_household_pks` doesn't either. The two functions you cite govern what a user sees and edits in the workspace, not what goes to HOPE.
    - Filtering would reintroduce the bug this PR fixes: a collector dropped from the queryset gets no `IND-…` tag, so `map_role_value` fails with `no mapping for primary_collector_id=…` and the households lose their collector.
    - The state isn't reachable, `removed=True` is only written by `_mark_rdp_beneficiaries_removed`, which marks households plus `household_id__in` members, and household-less collectors are excluded by construction.

2. Is this really the possible here during the kobo import?

3. This is already fixed as part of Vitali's comment.

4. fixed.

5. Agreed it's redundant, but fixing it cleanly needs structural changes in the validator. Leaving as-is for now

src/country_workspace/utils/import_flow/transformations.py
roma-valor: done

src/country_workspace/utils/import_flow/transformations.py
roma-valor: done

src/country_workspace/contrib/kobo/sync.py
roma-valor: done

PR #440, opened 2026-07-29
Feature/make rdi replacement asynchronous during rdp push

vitali-yanushchyk-valor:


src/country_workspace/rdp/push/repository.py
```diff
@@ -0,0 +1,84 @@
...
+        if (rdp := lock_rdp_push_attempt(rdp_id=rdp_id, push_attempt_id=push_attempt_id)) is None:
+            return None
+        if rdp.hope_rdi_id is not None:
+            return None
+        rdp.hope_rdi_id = "N/A"
+        rdp.save(update_fields=["hope_rdi_id"])
+        return rdp
+
+
+def start_rdp_push_attempt(*, rdp: Rdp) -> UUID:
```
arsen-vs: Maybe we can make this as method into Rdp class?

src/country_workspace/rdp/push/repository.py
```diff
@@ -0,0 +1,84 @@
...
+    """Start a new push attempt for an already-locked RDP."""
+    push_attempt_id = uuid4()
+    rdp.status = Rdp.PushStatus.PUSH_PENDING
+    rdp.push_attempt_id = push_attempt_id
+    rdp.save(update_fields=["status", "push_attempt_id"])
+    return push_attempt_id
+
+
+def lock_rdp_push_attempt(*, rdp_id: int, push_attempt_id: UUID) -> Rdp | None:
```
arsen-vs, line 41: We can create custom manager and add this method to manager? And just use Rdp.objects.lock_rdp_push_attempt(...)

src/country_workspace/rdp/push/repository.py
```diff
@@ -0,0 +1,84 @@
...
+        .filter(
+            pk=rdp_id,
+            status=Rdp.PushStatus.PUSH_PENDING,
+            push_attempt_id=push_attempt_id,
+        )
+        .first()
+    )
+
+
+def finish_rdp_push_attempt(*, rdp: Rdp, status: Rdp.PushStatus, hope_rdi_id: str) -> None:
```
arsen-vs: Method on Rdp instance?

src/country_workspace/rdp/push/repository.py
```diff
@@ -0,0 +1,84 @@
...
+    """Finish an already-locked RDP push attempt."""
+    rdp.status = status
+    rdp.hope_rdi_id = hope_rdi_id
+    rdp.push_attempt_id = None
+    rdp.save(update_fields=["status", "hope_rdi_id", "push_attempt_id"])
+
+
+def rdp_for_push(*, pk: int) -> Rdp:
```
arsen-vs, line 55: Let's maybe add to Manager this one?

src/country_workspace/rdp/push/workflow.py
```diff
@@ -0,0 +1,338 @@
...
+            rdp_id=rdp_id,
+            push_attempt_id=push_attempt_id,
+            hope_rdi_id=rdi_id_to_reset,
+        )
+        if isinstance(exc, RdpWorkflowError):
+            raise
+        raise RdpWorkflowError({"errors": [str(exc)], "rdp_id": rdp_id}) from exc
+
+
+def handle_push_ready_callback(*, rdp_id: int, push_attempt_id: UUID) -> bool:
```
arsen-vs, line 220: As option we can maybe do this in class based style? (not required)

src/country_workspace/contrib/hope/rdi/api.py
```diff
@@ -37,15 +60,18 @@ def _post(self, route: Route, payload: Any, *, rdi_id: str | None = None) -> dic
...
-            match exc.response.status_code:
-                case HTTPStatus.NOT_FOUND:
-                    return RdiDeleteResult.DELETED
-                case HTTPStatus.CONFLICT if exc.error_code == RDI_ALREADY_MERGED_ERROR_CODE:
-                    return RdiDeleteResult.ALREADY_MERGED
+            if exc.response.status_code == HTTPStatus.NOT_FOUND:
+                return RdiResetResult.NOT_FOUND
+            if exc.response.status_code >= HTTPStatus.INTERNAL_SERVER_ERROR:
+                raise HopeRdiResetUnconfirmedError(str(exc)) from exc
             raise
```
roma-valor, line 87: Retrying a push used to treat an already-merged HOPE RDI as success: CW skipped creating a new RDI and marked the RDP SUCCESS.
This PR dropped that path. reset_rdi() only handles 404 and 5xx. A 409 (or any other 4xx) now fails the push and leaves the RDP in FAILURE.
Please confirm this is intended

roma-valor:
one comment from me, otherwise looks good

src/country_workspace/contrib/hope/rdi/api.py
vitali-yanushchyk-valor: > Retrying a push used to treat an already-merged HOPE RDI as success: CW skipped creating a new RDI and marked the RDP SUCCESS. This PR dropped that path. reset_rdi() only handles 404 and 5xx. A 409 (or any other 4xx) now fails the push and leaves the RDP in FAILURE. Please confirm this is intended

Yes, you're right. The HOPE rdi_reset endpoint is not implemented yet, so I originally planned to add handling for these responses in a separate PR. In any case, I’ve added it to this PR now based on the preliminary response contract shared by Kamil.

src/country_workspace/rdp/push/repository.py
vitali-yanushchyk-valor: I’d keep this in the repository, since it contains push-attempt-specific semantics and row locking rather than being a general manager operation.

src/country_workspace/rdp/push/repository.py
vitali-yanushchyk-valor: I’d prefer to keep this in the repository for now since it’s specific to the push workflow rather than a general Rdp query.

roma-valor (approved):

